PR #2, opened 2016-06-02
fixed jl path

fischer1:
fixed chelpers path


maxhutch:
Looks great.  Thanks @fischer1

PR #3, opened 2016-06-03
Bgq mxm

maxhutch:
Adding bgq assembly version of mxm for common matrix sizes.  Also includes cmt_dummy, by accident.


maxhutch:
I'm merging this so I can test how well it resolve histories.

PR #12, opened 2016-06-07
Travis support with docker!

maxhutch:
Whale say yay!


maxhutch:
This actually works; see [here](https://travis-ci.org/maxhutch/nek5000/builds/136011101)

RonRahaman:
Looks good!  Super exciting, Max.

Issue #16, opened 2016-06-29
unported path in postnek's makefile


maxhutch:
[This line](https://github.com/Nek5000/nek5000/blob/develop/tools/postnek/makefile#L80) was only half ported: the blas dependency still points to nek/blas instead of core/blas

RonRahaman:
@maxhutch, is nek5000/3rd_party/ an artifact?  Should it be deleted?

maxhutch:
No.  That's a landing point for 3rd_party add-ons, e.g. KTH's toolbox.  It
should stay empty.

maxhutch:
Fixed in #22

PR #18, opened 2016-07-01
Updated makefile for postnek and removed unused directory

RonRahaman:
This pull request addresses issue #16 

postnek's makefile pointed to the "3rd_party" directory that existed in the SVN repo.  This directory was moved to core/3rd_party during the transition to Git.

Additionally, the top-level "3rd_party" directory was empty and unused.  Both users and developers were understandably confused.  I deleted that directory to avoid further confusion.

### DIFF
--- a/3rd_party/.gitignore
+++ b/3rd_party/.gitignore
@@ -1,4 +1,0 @@
-# Ignore everything in this directory?
-*
-# Except this file
-!.gitignore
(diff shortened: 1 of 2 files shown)


maxhutch:
The top level `3rd_party` was deliberate as a consistent landing spot for 3rd party code, and was discussed with KTH et al.  For example, if I write and automatic download/install into makenek for LIBXSMM, I'll download it to there.  If anything, I think we should consider getting rid of `core/3rd_party`, but in a different PR.

RonRahaman:
Thanks, Max.  Didn't realize this was part of the discussions with KTH.  I'll close this PR and submit another with just the changes to `tools/postnek/makefile.`

PR #19, opened 2016-07-12
Cray patch

nicooff:
Tag for sending and receiving changed from global (eg) to local (e) element number to avoid problems with MPI_TAG_UB on Cray.

Patch from P. Schlatter.


maxhutch:
Thanks, @nicooff

stgeke:
It looks like this patch is not sufficient!
We are still tagging some messages with ielg (e.g. in bin_rd1_mesh).

RonRahaman:
Maybe we can make a new issue referencing @nicooff's bugfix?  Seems like this would benefit from some more discussion

pschlatt1:
It might be that we did not use some of these other routines to make it work on the cray. I can look at again in a few days. But perhaps with an issue it is best.

stgeke:
Yes, let's open up a new bug issue. 

So you use ASCII mesh data instead of binary (.re2 file) for your >2M element computations on the Cray?

pschlatt1:
no we don't. I cannot say what the problem is right now, but let me check later.

stgeke:
Thanks @pschlatt1 
Sorry for being pushy. There is a scaling workshop at ANL (testing the new Cray system) today and @ggiannako is trying to get some results. Unfortunately the code bails out because of invalid message tags.

pschlatt1:
ok, I see. Then I will try to look at it asap and see if I can see something.

pschlatt1:
I found the problem; the changes in connect2.f were missing in the original PR. @nicooff will commit it asap.

RonRahaman:
@pschlatt1 and @nicooff, when you submit your PR, please merge into Nek5000:release/v16.0.0 instead of Nek5000:develop.  We're in pre-release right now.

nicooff:
Hi everyone,

I apologize for the delay. I did a pull request with the changes into
Nek5000:release/v16.0.0.

Nicolas

PR #20, opened 2016-07-12
Thanks Dillon!

maxhutch:
`--single-branch` doesn't work on MCS machines.


maxhutch:
Based my branch on master instead of develop.  Sorry.

PR #26, opened 2016-07-14
Redirecting move warnings to /dev/null.  Closes #25

### LINKED ISSUE #25: Remove extra 'error' messages from run scripts
Many of the run scripts have `mv` commands to back-up output files, e.g. [nekmpi:6](https://github.com/Nek5000/Nek5000/blob/develop/bin/nekmpi#L6):

```
mv $1.log.$2 $1.log1.$2
```

When the output files aren't already present, these produce error messages that aren't actually errors.  This is confusing to users.  We should wrap those moves in if statements conditional on the existence of the file that is being moved.

maxhutch:
I didn't do the arguably more proper:

```
if [ -e <file> ]; then
  mv <file> <file1>
fi
```

because it wouldn't work in all shells, e.g. csh.

### DIFF
--- a/bin/nek
+++ b/bin/nek
@@ -1,8 +1,8 @@
 echo $1        >  SESSION.NAME
 echo `pwd`'/' >>  SESSION.NAME
 rm -f ioinfo
-mv -f $1.log $1.log1
-mv -f $1.sch $1.sch1
+mv -f $1.log $1.log1 2>/dev/null
+mv -f $1.sch $1.sch1 2>/dev/null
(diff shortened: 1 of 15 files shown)


stgeke:
@RonRahaman Any objections against this change?

RonRahaman:
This should be merged into release/v16.0.0, not develop.  I'll open a new PR, then you can feel free to merge.

RonRahaman:
Replaced with PR #67

PR #35, opened 2016-07-20
Msh2nek

nicooff:
Added the converter msh2nek to the tools folder. 

Msh2nek is a Python utility for converting a Fluent mesh (.msh) to a .rea file that can be read by Nek5000. It was initially developed by Mikael Mortensen at University of Oslo. Further modifications were implemented by Jing Gong, Prabal Negi and Ricardo Vinuesa at the Royal Institute of Technology (KTH).

### DIFF
--- a/tools/msh2nek/README.md
+++ b/tools/msh2nek/README.md
@@ -1,0 +1,25 @@
+# MSH2NEK
+
+Msh2nek is a Python utility for converting a Fluent mesh (.msh) to a .rea file that can be read by Nek5000. It was initially developed by Mikael Mortensen at University of Oslo. Further modifications were implemented by Jing Gong, Prabal Negi and Ricardo Vinuesa at the Royal Institute of Technology (KTH).
+
+
+## NOTES
+- Supports only hex8 format.
+- Curvature can be introduced at the domain boundaries by the user.
+- A detailed description of functionalities can be found in the Python file.
+
+## To get started:
+Starting from a file 'msh_file.msh', the script can be used as follows:
+
+'''
+Unix> ipython
+
+...
+[1]: from mshconvert import *
+[2]: convert('msh_file.msh',bcs={10:'W',11:'v',12:'v',13:'O'},curves={12:{'type': 'm'})
+'''
+
+Inputs:
+- The first argument is the .msh file containing the mesh. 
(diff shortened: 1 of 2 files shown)


stgeke:
I really encourage contributions like this!

However, this pull request brings up a general question: Do we want to add all available user contributions to the official project?

I think the answer to this question is yes and no. Yes, if the code meets our standards (whatever that is) and we believe it’s of general interest. This requires a review by an experienced maintainer to see what’s needed for integration.  This can become tedious procedure.

To minimize the effort for sharing contributions, it would be great to have a dedicated area (like a bazar) e.g. a wiki page listing all the contributions including a short description and the link for download. 

What are your thoughts on that?

oanaoana:
I agree that all submissions would clutter the folders in time, however the git was designed with empty folders especially for such future contributions, and tools is a suitable place for mesh converters (as would be 3rd party which stands empty). However since the design was to be an encompass all and not a two repos: main and dependencies there is no other way around it. 

Furthermore the spot for such things used to be on the homepage in meshes. I do not see how the wiki could be referenced properly on the homepage though it could and it should. This submission is similar to the exo2nek and both would fit in the same place.

RonRahaman:
In this case, we had discussed @nicooff 's contribution at our weekly meetings, so many of the core developers had reached a consensus about including it.  @stgeke, I'm sorry that the discussion didn't make it to GitHub for the other off-site developers.

> This requires a review by an experienced maintainer to see what’s needed for integration. This can become tedious procedure.

You're totally right, the approval process can get tedious.  However, I think 3rd-party tools submissions will be relatively infrequent and will not interfere with the functioning of the core code.  Therefore, the necessary discussions shouldn't block overall development.    

> To minimize the effort for sharing contributions, it would be great to have a dedicated area (like a bazar) e.g. a wiki page listing all the contributions including a short description and the link for download.

We do actually have a few pages on the WordPress site for stuff like this ([for example](https://nek5000.mcs.anl.gov/meshes/)).  However, they're really hard to find if you don't know where they are.  Definitely in need of improvement.  

Any ideas about how to make them easier to find?  Maybe this also pertains to this issue: Nek5000/NekDoc#12

stgeke:
Ah ok, didn't know that. Sure, the core-developers decide what to add.

PR #36, opened 2016-07-20
Making the interpolation higher-order consistent

vijaysm:
The original NEK interpolation used for coupled physics was second order at best (spatially). This was due to an approximation to store the injection of the higher order vertex DoFs alone and use an element average for the rest. Instead, now we make use of NEK's native projection routines to go from a high-order GLL basis to either linear/quadratic Lagrange basis elements exactly. This means the coupled solution transferred to the other physics should be much more accurate than before!


RonRahaman:
This pull request was made in error when @vijaysm and I were working on his fork.  Please disregard.

PR #37, opened 2016-07-21
Updated Travis badge after renaming Nek5000 repo

RonRahaman:
The image for the Travis badge isn't loading.  I think it's because we changed the name of the repo (nek5000 -> Nek5000) and Travis doesn't redirect.  This fixed it in my fork.


maxhutch:
Ah, case sensitivity.

PR #38, opened 2016-07-26
Bug fixes for exo2nek

ggiannako:


ggiannako:
I have worked with Stefan on some issues and we believe it should work now.

PR #48, opened 2016-08-05
Changing line continuations.

### LINKED ISSUE #47: Code does not compile on BG/Q with the PPLIST="BGQ" flag
The code does not compile on the BGQ with the PPLIST="BGQ" flag.
The mxm_wrapper.f and mxm_bgq.f contain f90 syntax. The code cannot compile neither with mpixlf90 nor with mpixlf77.

Even if the syntax in mxm_wrapper.f and mxm_bgq.f is converted to f77, we get “undefined references” to some blas routines at compile time.
The problem is not fixed either with the BLAS_MXM flag of PPLIST.

Are there any other flags that should be used along with "BGQ"?
If yes, it would be good to have it documented.

maxhutch:
This should fix #47

### DIFF
--- a/core/makenek.inc
+++ b/core/makenek.inc
@@ -378,7 +378,7 @@ if [ $? -eq 0 ]; then
 fi
 echo $PPLIST | grep 'BGQ' >/dev/null 
 if [ $? -eq 0 ]; then
-   MXM_USER="mxm_std.o mxm_bgq.o " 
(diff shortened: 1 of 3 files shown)


stgeke:
Reverted my merge because I think it should go into the release/v16.0.0 branch.
@RonRahaman Am I wrong?

RonRahaman:
Yes, @stgeke, you are correct.  I see that you reverted in #60.  Thanks for that!  I submitted a new PR for this in #61

PR #53, opened 2016-08-08
Synching up short_tests/lib with NekExamples/lib

RonRahaman:
The library for short_tests/NekTests.py is being synched up with the library from NekExamples.  It features verbose output for makenek, maketools, genmap, and genbox.


RonRahaman:
Approving myself b/c of time constraints.

PR #54, opened 2016-08-09
Deleted axi.rea, included axi.box in short_tests

RonRahaman:
In NekTests.py, both genbox and genmap are run. Previously, genbox failed because axi.box wasn't present.  However, the whole test succeeded because axi.rea was already present.  

Now, axi.rea is generated when genbox is run.  This should provide a better test for genbox.


RonRahaman:
Closing this PR, then reopening from from a bugfix branch

PR #73, opened 2016-08-25
makenek appends $FFLAGS to FORTRAN compiler flags

RonRahaman:
The FFLAGS variable is used in GNU make and autotools.  In a GNU build system the user may set FFLAGS in their environment to set additional compiler flags.  The build system will honor whatever the user has set in FFLAGS, in addition to whatever is required by the build system.  

Setting FFLAGS is a very clean way to extend functionality of makenek without changing the source files.  For example, the user can use FFLAGS to set the memory model, if their problem or computing platform requires it.

### DIFF
--- a/core/makenek.inc
+++ b/core/makenek.inc
@@ -210,23 +210,23 @@ fi
 
 # assign F77 compiler specific flags
 case $F77comp in
-  *pgf*)        P="-r8 -Mpreprocess"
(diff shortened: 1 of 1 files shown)


stgeke:
I like your idea but let's do it right and switch to the GNU default variables (CC, FC, CFLAGS, FFLAGS, ...).

RonRahaman:
Glad you're interested.  I'll spin this off into an issue and close this PR.

Issue #68, opened 2016-08-16
Code does not compile with implicit=none


stgeke:
To fix this, we need to explicitly declare all variables used in all `INCLUDE` files. 

I suggest to add a `SIZE.usr` (this file will live in user space). `SIZE`  (static file in the source) contains whatever we want to hide from the user (e.g. variable declarations or other internals). `SIZE` will include `SIZE.usr`. This is the cleanest solution I can think of and it improves the user experience (`SIZE` is less messy). 

No code changes are required. However, the user needs to use the new `SIZE.usr` template (distributed with the source). I never liked the `makenek` magic to tweak old `SIZE` files. 

Overall I do _not_ think this is a big deal as we do something similar with `makenek`.

We had a similar discussion at the user meeting.
@fischer1 @pschlatt1 What are your thoughts on that?

fischer1:
The key step is to embed the declarations in all the INCLUDE files.

I'm not 100% certain why we need SIZE.usr ?   Can you remind me?  I recall
something like this from before --- Oh yes -- Now I know.   We need to declare
all the SIZE arrays too.

My suggested remedy:
1.  Put declarations in SIZE for all the Examples.
2.  Any existing SIZE files out there are either:
   
   (A)   They are OK because the user in question is not going to
           run IMPLICIT NONE
   
   (B)   The user in question IS going to run IMPLICIT NONE and will
           either:
   
   ```
       B.1 -- Grab a SIZE file from an example, or
       B.2 -- Modify on his own in any case.
   ```

To me, the scenarios (1) and (2) cover all SIZE issues and we don't need
an extra include into SIZE  [ unless you want to do other things in the SIZE.usr
file, such as the other appends that we currently have ].

I'm not that unhappy with what we have, for the following reason:

```
When I look at my SIZE file and see those declarations, it reminds of what
options Nek offers that I'm not currently using.

If those declarations are hidden in SIZE.usr, which I might never edit, then
I'd not be reminded of those options.
```

Paul

stgeke:
I am ok with your suggestion although I would prefer a SIZE.usr (containing all available parameters + optional user stuff) and SIZE (containing all declarations).

RonRahaman:
@stgeke, if you're going in this direction, then we shouldn't target the change for v16.0.0.  We should wait for the next major revision (v17.X.X) to change the interface (see http://semver.org/)

stgeke:
I think we go with Paul's suggestion.

[...]

PR #79, opened 2016-09-02
Issue 68

nicooff:
- Variables declaration has been added to all the include files. A file SIZE.template, which is a SIZE file with the variables declaration, has been added to the core folder. This file can serve as a basis to update the SIZE files in the examples (but I guess it should not remain in core in the long term). This means it is now possible to compile with 'implicit none'.
- We chose the following layout for declaration:
  
  real a(na)
  common /aa/ a
  
  integer b(nb)
  common /bb/ b
- One issue still pending is that several variables with the same name are declared in different include files:
  - x in GMRES, NEKUSE and ORTHOSTRS
  - mu in GMRES and NEKUSE
  - dd in FDMH1 and SEMHAT
  - zgl in SEMHAT and WZF


stgeke:
I tested it and it looks good to me. 
@fischer1 Merge?

stgeke:
Fixes #68

PR #81, opened 2016-09-03
Fixed inclusion of MPI in PPLIST definition when IFCMT==false

RonRahaman:
@fischer1 and his students discovered an error in makenek when compiling w/o MPI.  In particular, the MPI proprocessor symbol is always defined when IFCMT==false.  This overrides the definition of IFMPI in makenek.  

This PR fixes that issue.


fischer1:
Thanks Ron!

PR #93, opened 2016-09-16
bugfixes in local_grad2_t, navier1.f

hackljf:
gs_setup should not have an integer*8 for that particular argument. Even though ut isn't used, it's never good to have arrays dimensioned with variables unless they're dummy args. This fix prevented NaNs in gfortran+openmpi SMP runs.


fischer1:
Thanks for the local_grad2_t fix Jason.

PR #99, opened 2016-09-27
Support for 'o' and 'on' boundary conditions added to non-stress formulation.

mattiasbr:
Enables the use of inhomogeneous 'O' and 'ON' boundary conditions (i.e. 'o' and 'on') of the form
nu dv/dy - p = pa
nu du/dx - p = pa
for non-stress formulation, where pa may be specified in userbc().

### DIFF
--- a/core/bdry.f
+++ b/core/bdry.f
@@ -385,6 +385,12 @@ C
          do 50 iel=1,nelt
          do 50 iface=1,nfaces
             cb=cbc(iface,iel,ifield)
+            if ((cb.eq.'o  ' .and. IFSPLIT) .or.
+     $          (cb.eq.'on ' .and. IFSPLIT)) then
+               if (nid.eq.0) write(6,*)
+     $              "Error: BC 'o' and 'on' not supported for PN-PN."
+               call exitt
+            endif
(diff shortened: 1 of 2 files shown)


pschlatt1:

fischer1:
@ananiast Does this work also for pn/pn ?  Or do we need to make an error check
if this BC is used for pn/pn ?   

@mattiasbr Do we have an example by which we can test this feature ?  - Paul

ananiast:
It won't be enough for Pn-Pn because we need to modify some other routines as well,
but it should be OK for Pn-Pn-2. I suggest we accept it but you should add an error check
about this

stgeke:
@ananiast should we merge this pull request?

stgeke:
@mattiasbr Please resolve the conflicts. Thanks!

mattiasbr:
The conflicts are now resolved and an error check for PnPn has been added to bcmask().

pschlatt1:
Can we merge it?

stgeke:
@ananiast Did you implement this change with another pull request?
In this case we can close the request w/o merging.

PR #101, opened 2016-10-03
Added new SIZE template

stgeke:
- Use implicit none in lowMach_test
- Updated template for .usr file

### DIFF
--- a/core/makenek.inc
+++ b/core/makenek.inc
@@ -8,15 +8,14 @@ echo "makenek - automatic build tool for Nek5000"
 if [ "$PPLIST" == "?" ]; then
   echo "available pre-processor symbols:" 
   echo "  MPIIO     use MPI-IO I/O kernel (experimental)"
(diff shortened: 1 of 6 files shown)


stgeke:
@RonRahaman Any idea why the test cases w/o MPI fail? I don't think this is specific to this pull request.

core/makenek.inc
```diff
@@ -251,6 +250,7 @@ if [ "$IFCMT" == "true" -o "$IFCMT" == "yes" ]; then
 else
   # default
   IFCMT=false
+  PPLIST="${PPLIST} MPI"
```
RonRahaman: Upstream changes from #81 weren't included here.

RonRahaman: To be more precise.  Nek5000:release/v16.0.0 includes the fix from PR #81, but stgetke:feature/SIZE clobbered that fix.

PR #102, opened 2016-10-06
Added circular-cartesian 2d mesh capability

kmittal2:
I have added a circular-cartesian mesh option under Curve Sides for 2D geometries. This allows the user to produce meshes external to a circle. It can be used to produce meshes for annulus or for circular-to-cartesian grid (similar to meshing hole in a box).


fischer1:
done

PR #111, opened 2016-10-20
corrected dsset initialisation in setup_topo

adampep:


pschlatt1:

Issue #109, opened 2016-10-18
Time integration order


adampep:
Nek uses two parameters to control order of time integration NAB and NBD (in TSTEP). Max value of NBD can be set by parameter(27) (general:tOrder), however max value NAB is hard-coded to be 3. For optimal forcing we have to restrict time integration to Euler and we would like to be able to control both NAB and NBD. Simple solution would be to keep current behaviour for positive values of parameter(27) and to set both parameters to NBDINP for negative one. Is it OK?

stgeke:
Resolved in #112

PR #112, opened 2016-10-21
Time integration order #109

adampep:
negative value of param(27) sets max of nab to nbdinp


pschlatt1:

Issue #128, opened 2016-10-30
mean pressure in Pn-Pn low Mach


fischer1:
@ananiast should we append mean pressure to end of field file ...

stgeke:
As discussed, it should go into the header. Will this break the Visit reader?

pschlatt1:
apparently, this change causes problems restarting from old files without that variable. Perhaps one could fix that when reading the internal file/header.

stgeke:
Please file a bug (issue) including some instructions how to reproduce it. Thanks.

fischer1:
@stgeke @pschlatt1 

I just filed a bug report and posted a pointer to an exmple.

pschlatt1:
Great! The fix is not very difficult; one should just make the "parser" of the internal file more robust. I guess we could look at it at the beginning of next week.

stgeke:
Implemented in #134

PR #134, opened 2016-11-03
added p0th to hdr for restarts

stgeke:


stgeke:
Fixes #128

stgeke:
@RonRahaman One test failed because of the runTime check. I don't like this check in general because our test environment is not deterministic.

PR #137, opened 2016-11-03
Updates to short_tests

RonRahaman:
This PR introduces a number of updates to short_tests:
* When using the NekTests.py, the SIZE files for short_tests are generated from SIZE.template
* Each example in short_tests also contains a SIZE file based on SIZE.template, should any users want to run standalone Nek5000.   
* Users may run NekTests.py from the short_tests/ directory without setting any environment variables.  The defaults correspond to the current directory structure but may still be overriden; see short_tests/README for more info
* Users may run NekTests.py as a standalone executable with command-line arguments instead of env variables.  For example `./NekTests.py --f77 pgif77 --cc pgicc --ifmpi false`.  See short_tests/README for more info

### DIFF
--- a/short_tests/NekTests.py
+++ b/short_tests/NekTests.py
@@ -1003,9 +1003,47 @@ class CylRestart_Pb(NekTestCase):
 
 class Eddy_EddyUv(NekTestCase):
     example_subdir  = 'eddy'
-    case_name        = 'eddy_uv'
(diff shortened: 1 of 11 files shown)


RonRahaman:
Also omitted tests for total solver time, as discussed with @stgeke

RonRahaman:
Merged after offline discussions w/ @stgeke

PR #139, opened 2016-11-04
Continuous merge of release/v16.0.0 into develop

RonRahaman:
We are merging the current state of release/v16.0.0 into master so users of master can benefit from the extensive bugfixes.  You can think about this as a release candidate prior to the stable release.

cc: @stgeke


stgeke:
@RonRahaman Can merge in the latest changes of release/v16.0.0?

Issue #88, opened 2016-09-15
(linear) Adjoint equations for temperature


pschlatt1:
extend the current linear adjoint for velocity with the respective terms in the temperature equation (with Boussinesq approximation). Question: Should we contribute that already for R16?

stgeke:
Can somebody please update me on this otherwise I am closing this issue for now.

pschlatt1:
yes, can be closed. The relevant code was in PR #141 and the corresponding example in #337.

PR #141, opened 2016-11-06
Add adjoint for passive scalar (just T) #88

cliosaglietti:
advabp_adjoint (in perturb.f): has an extra term to be considered if ifheat is active.
convabp (in perturb.f): has one term less if ifadj is true. 
ADJOINT: has extra variables.

The changes in the code validated with Arnoldi (spectrum of adjoint and direct eigenvalues matching).


pschlatt1:
Is there any objection against merging this one?

PR #143, opened 2016-11-08
2D smoother in prenek

kmittal2:
I have added a 2D smoother in prenek.. it can be accessed by clicking :
global refine option -> Smooth (this brings up options, including an undo option) -> Smooth (this runs the smoother after the user provides the number of iterations of laplacian and optimization smoothing the user would like to run).


stgeke:
Please merge into release/v16.0.0 not develop

PR #158, opened 2016-11-16
Cvode build

RonRahaman:
This introduces regression tests for the moving cylinder case using cvode.  

In NekTests.py, I've added two new tests
1. PnPn parallel with 1e3 timesteps
2. PnPn parallel with 1e4 timesteps

.travis.yml has been updated to run the 1e3 timestep case.  The 1e4 timestep case exceeds the maximum allowable time limit for Travis jobs.  It can still be run with Jenkins (TODO).


RonRahaman:
Changes are limited to NekTests and .travis.yml.  I'm approving now so we can merge these upstream changes into ufccmt/cmt-merge-2

PR #165, opened 2016-12-09
update prenek/glomod.f to include missing col2 subroutine

klheisey:


stgeke:
@RonRahaman Why does the tests fail?

RonRahaman:
Looks like the two serial tests fail because of the bug we fixed in #164

RonRahaman:
@klheisey, if you pull in upstream changes from Nek5000:release/v16.0.0, then the tests should pass

PR #173, opened 2016-12-15
added prenek .par/re2 input file support

klheisey:


stgeke:
Why do we need to read the .par file?

klheisey:
to read in the logicals (ifaxis, iftran,ifmvd, ifadvc) but they could also be set to false and assumed that the user will adjust during mesh construction. I think there are just a few menu options that were set to display only when certain logicals were True. I'll adjust and let you know

stgeke:
@klheisey I think it would be good to remove this dependency. What do you think?

klheisey:
yea, I think that is good. I am working on the details, but I will remove these dependencies.

emerzari:
While removing the dependency may be good, there is also some "downside". For instance: If the users forgets to change ifheat to true, he/she will not be prompted to insert bcs for the temperature equation. Right now I typically rely on the base .rea I start from to do that.

It's  a slight change of workflow, and may not be a bad one, just making sure everyone is aware and we properly document this change.

stgeke:
I recommend to setup boundary conditions in usrdat2 at runtime. In this case I use the v-BC in .re2 as a template or boundary patch identifier.

On 20 Dec 2016, at 23:31, emerzari <notifications@github.com <mailto:notifications@github.com> > wrote:

While removing the dependency may be good, there is also some "downside". For instance: If the users forgets to change ifheat to true, he/she will not be prompted to insert bcs for the temperature equation. Right now I typically rely on the base .rea I start from to do that.

It's a slight change of workflow, and may not be a bad one, just making sure everyone is aware and we properly document this change.

emerzari:
Stefan, that's very good for advanced user and I do that all the time for instance.
However, I think that the standard workflow should have  setup of the BCs at the meshing stage (e.g., as an input).

For very complex geometries it is also very hard to modify at runtime as some regions may require special handling.
Moreover, although your suggestion to use the velocity BCs as a template usually works for fluid-only simulations, for CHT calculations it is not always true.

stgeke:
Some what do you suggest for prenek?

On 21 Dec 2016, at 08:58, emerzari <notifications@github.com <mailto:notifications@github.com> > wrote:

Stefan, that's very good for advanced user and I do that all the time for instance.
However, I think that the standard workflow should have setup of the BCs at the meshing stage (e.g., as an input).

For very complex geometries it is also very hard to modify at runtime as some regions may require special handling.
Moreover, although your suggestion to use the velocity BCs as a template usually works for fluid-only simulations, for CHT calculations it is not always true.

stgeke:
@klheisey What is needed to make this change happen?

PR #206, opened 2017-01-26
Serial AMG setup using Hypre (Issue-205)

nicooff:
A serial version of the AMG setup using Hypre can be found in tools/amg_hypre.

### DIFF
--- a/tools/amg_hypre/README.md
+++ b/tools/amg_hypre/README.md
@@ -1,0 +1,44 @@
+============================
+SERIAL AMG SETUP USING HYPRE
+============================
+
+Serial version of the AMG setup for Nek5000 using Hypre, a linear algebra library (http://computation.llnl.gov/projects/hypre-scalable-linear-solvers-multigrid-methods). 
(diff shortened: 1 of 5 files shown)


stgeke:
I recommend two small changes:
Can you change the makefile to make it compatible with maketools. Please modify `tools/makefile` such that this tool does not get compiled by default (similiar to exo2nek). Otherwise maketools will fail if Hypre is not installed.

nicooff:
I fixed maketools and added CFLAGS_IN to the amg_hype/makefile. And no more capital letter to makefile.  Are these changes sufficient?

stgeke:
Not really as the makefile is should use the maketools variables (CC, LFLAGS_IN, ...). Not a big deal, I can take care of this. 

Why is code using MPI if only one rank is supported?

stgeke:
Merged in #209

nicooff:
Thanks!

The Hypre library requires MPI. I am currently not using the parallel features because of matrix partitioning and because the smoother is only serial for the moment.

stgeke:
You can build HYPRE w/o MPI.

On 31 Jan 2017, at 10:51, Nicolas <notifications@github.com <mailto:notifications@github.com> > wrote:

Thanks!

The Hypre library requires MPI. I am currently not using the parallel features because of matrix partitioning and because the smoother is only serial for the moment.

nicooff:
Right. Thank you.

PR #214, opened 2017-01-31
Issue hpf

negips:
High-pass filter implementation added to the core.
Uses param space for input variables.
param(110-112)
Added parameters under GENERAL section for .par file.


stgeke:
This is a great new feature. Thank you!

It looks like the source terms you add are fully explicit. Will this introduce an additional timestep restriction. What is your experience?

We decided some time ago that turbulence models should live in usr-space and not in code. This approach is less intrusive and requires no modification of the source code. 

In your case a user would do the following: 

- Add `#include les_hpf.h` to the top of your .usr file (this way the code gets compiled)
- In userf()   `ffx = les_hpf(vx)`, similar for `ffy` and `ffz`
- In userq()   `qvol = les_hpf(T(1,1,1,1,ifield-1))`
- In usrdat3() `les_hpf_setup(addmodes, kai, ifboyd)`

I am happy to accept this pull request if you send me your `les_hpf.h` and `LES_HPF` (providing the functionality described above) including a short_test for our regression test suite.

stgeke:
@pschlatt1 What is the correct name for this model? HPF-RT?

pschlatt1:
I would typically call it the ADM-RT model (Ref. [Schlatter, Stolz, Kleiser, IJHFF 2004](http://www.sciencedirect.com/science/article/pii/S0142727X04000281)), since it was originally based on the stabilisation term of the ADM model. ADM at that point referred to having the same implicit filters of the original ADM implementation. But of course, now the filters are very different, so perhaps HPF-RT (i.e. saying high-pass filtered relaxation term) might be quite appropriate.
As a side note, this type of modelling is very similar to SVV (spectrally vanishing viscosity), with the only difference being the shape of the filter). But the RT approach appears to be more general as we do not base it on a 2nd order term.

stgeke:
Ok lets go for HPF-RT.
Out of curiosity: How do you pick the relaxation constant? Why is the current implementation not based on a dynamic procedure?

On 1 Feb 2017, at 14:06, Philipp Schlatter <notifications@github.com <mailto:notifications@github.com> > wrote:

I would typically call it the ADM-RT model (Ref. Schlatter, Stolz, Kleiser, IJHFF 2004 <http://www.sciencedirect.com/science/article/pii/S0142727X04000281> ), since it was originally based on the stabilisation term of the ADM model. ADM at that point referred to having the same implicit filters of the original ADM implementation. But of course, now the filters are very different, so perhaps HPF-RT (i.e. saying high-pass filtered relaxation term) might be quite appropriate.
As a side note, this type of modelling is very similar to SVV (spectrally vanishing viscosity), with the only difference being the shape of the filter). But the RT approach appears to be more general as we do not base it on a 2nd order term.

negips:
@stgeke  There's a bunch of questions here which I'll try to answer with some background on why we tried to put this into the core rather than through the usr space.

First the relaxation term (HPF-RT as we are calling it here) is implemented to provide the same functionality as the filtering stabilization, which is already implemented, except that it does it via adding a forcing term instead of doing an explicit filter of the solutions at the end of the time step. This has a few advantages, the primary one being that the final solution preserves the divergence free criterion. While if you do an explicit filter of the solutions this divergence free quality of the solution is lost. In our tests, the loss can reach about 5 orders of magnitude. So if the pressure step solution was obtained to a tolerance of 1E-8, the explicit filtered solutions can have a divergence of 1E-3. Which is a very steep price to pay. Doing it via a forcing with a HPF-RT avoids this loss, since the solutions are not touched after the pressure step.
It has a secondary advantage in terms of its dissipative nature. The explicit filter acts on a modified basis functions so as to preserve element boundary values, which preserves C0 continuity. This change of basis results in the explicit filter not being purely dissipative and can sometimes add energy to the flow. In the HPF-RT formulation, such a change of basis is not required, and thus the added forcing is purely dissipative in nature.

So in that sense, the implementation is not meant as a turbulence model but rather as a (better) stabilization tool. That's the primary reason why we wish to include it into the core. That will also tell you why there is no dynamic procedure involved for the kai parameter. The implementation is modeled after the explicit filtering procedure. Hence the kai value is static and controlled by param(111). Of course the procedure can be leveraged to be used as a turbulence model should one desire, just as the current filter is sometimes used as an LES filter.

I wanted to include some documentation on how to choose kai, but I wasn't quite sure where to put it. To compare it with the current filter, a typical "wght" value we choose for stabilizing is 0.02 with 0 additional modes. An equivalent value of kai for HPF-RT procedure will be 0.02/deltaT (deltaT being the time-step), with 0 additional modes.

In terms of time-step restriction, the added forcing makes the flow unstable if kai ~ 1/deltaT. The exact value where it goes unstable depends on the polynomial order N. for N=16 the value is about 0.8/deltaT. But those values are very far away from the typical values one would use for stabilization. Hence we do not think it poses a time-step restriction under normal usage.

Almost all the results/numbers I mention here were presented in the Nek user meeting in Boston last year. I might still have the presentation lying around somewhere should you wish to see it.

So, long story short, HPF-RT is implemented for stabilization and not as a turbulence/les model. It does a much better job than the current explicit filter used by Nek and hence we think it should be part of the core code rather than user added module/feature.

stgeke:
Got it. For now, please implement it in the recommended way. We can think about a core integration as the feature gets more mature. Ok?

stgeke:
We want to continue to support the legacy SIZE. Please ensure that your implementation works for both versions.

pschlatt1:
ok, thanks for the discussion. Let me just make sure that we understand:
- is the idea of making Nek5000 modular via putting Fortran code in header files (e.g. les_hpf.h) and include them as needed via "#include"?
- keeping the support for legacy SIZE files essentially means that no new code (or any Nek5000 subroutine) can contain IMPLICIT NONE?
Thanks.

stgeke:
I wouldn't go that far. It's more the natural extension of what we did in the past where we introduced a new feature through an example (if possible). Include files will allow us to use/test the new feature in other cases easily. We also evaluate different options to make the code more modular but that's a different topic.

Yes, that's the implication unless we drop the legacy support or come up with a robust way adding the missing decelerations (for all existing SIZE flavors) to SIZE.

On 2 Feb 2017, at 13:50, Philipp Schlatter <notifications@github.com <mailto:notifications@github.com> > wrote:

ok, thanks for the discussion. Let me just make sure that we understand:

*	is the idea of making Nek5000 modular via putting Fortran code in header files (e.g. les_hpf.h) and include them as needed via "#include"?
*	keeping the support for legacy SIZE files essentially means that no new code (or any Nek5000 subroutine) can contain IMPLICIT NONE?
Thanks.

RonRahaman:
I've been sitting on a converter for an arbitrary SIZE file to the current format.  It's simple (< 50 lines of Python, delta my copious whitespace) and robust.  If I fixed it up, would that changer your decision to keep or scrap support for old-style SIZE files?

The in-progress converter is here (right now, it's just a parser; turning it to a converter is trivial):
https://github.com/Nek5000/Nek5000/blob/develop/short_tests/lib/parseSize.py

stgeke:
Sure, we can drop the legacy support if the user doesn't have to deal with it. Thanks Ron!
Please make sure that it really works for the different SIZE version out there. Not sure if the current NekExamples cover the whole zoo.

fischer1:
I guess script does not _FORCE_ implicit none, though, correct ?    Thanks!   :)

stgeke:
@fischer1 No, of course not!  It is just to make the compiler happy when somebody is using`implicit none` in his code.

RonRahaman:
The script is available in PR #217.  I'd especially like to hear your opinion regarding the interface.

stgeke:
A converter provides a reasonable solution to deal with `implicit none` and legacy SIZE files. If the tool becomes available (#217) we are good to go.

I can follow your argument that HPF is a core component as it is an alternative implementation of an existing stabilization mechanism. A potential SGS model (based on HPF) should live in usr-space i.e. no calls outside usr-routines and gets compiled on demand. 

We can merge it if you address the following:
* Can you use some of the existing scratch CBs for RTFILrv to lower the memory footprint?
* Some F77 compilers don't like comments starting with `!` in the 1st column

negips:
Thanks. I'll change the !s and use the common blocks. 
I'll make another pull request once I've tested things.

stgeke:
No need to open a new pull request. Just push to the existing one.

PR #215, opened 2017-02-01
Add k-omg turbulence model

stgeke:


stgeke:
@ananiast Please provide a short_test for our regression tests.

PR #223, opened 2017-02-12
WIP: various fixes

stgeke:


stgeke:
@RonRahaman It looks like the Pyhton scripts parse the makefile created by makenek. This PR changes some variable names like F77->FC, P->FFLAGS, lFLAGS->LDFLAGS. I guess that's why the regression test fail. What do you think?

RonRahaman:
@stgeke, yes, that would be the reason.  I can make a fix to NekTests.py then submit a PR to stgeke:master.  After you approve that PR, the changes will appear here.  Does that sound okay?

core/makenek.inc
```diff
@@ -88,99 +84,56 @@ if [ ! -f $SOURCE_ROOT_CORE/makefile.template ]; then
   exit 1
 fi
 
-# test F77 compiler
-which `echo $F77 | awk '{print $1}'` 1>/dev/null
-if [ $? -ne 0 ]; then
-  echo "FATAL ERROR: Cannot find $F77!"
+# test if F77 compiler exists
+if ! which -s `echo $F77 | awk '{print $1}'` ; then
```
RonRahaman: The `-s` option is valid for the `which` command in whatever flavor of BSD UNIX that Mac OS/X uses.  It is NOT valid for the flavor of GNU/Linux that Ubuntu and Debian use.  This will cause the `if` statement to fail and makenek to exit with the "FATAL ERROR: Cannot find .." error.  

This is what `man which` shows on my Ubuntu workstation:
```
WHICH(1)                                                                                                              General Commands Manual                                                                                                             WHICH(1)

NAME
       which - locate a command

SYNOPSIS
       which [-a] filename ...

DESCRIPTION
       which  returns  the  pathnames of the files (or links) which would be executed in the current environment, had its arguments been given as commands in a strictly POSIX-conformant shell.  It does this by searching the PATH for executable files matching
       the names of the arguments. It does not follow symbolic links.

OPTIONS
       -a     print all matching pathnames of each argument

EXIT STATUS
       0      if all specified commands are found and executable

       1      if one or more specified commands is nonexistent or not executable

       2      if an invalid option is specified

Debian                                                                                                                      1 May 2009                                                                                                                    WHICH(1)
```

RonRahaman (commented):
I have a fix for the issue in nekTests.py, but it exposed this issue with makenek.inc.  If I omit the -s flag, the tests succeed.  How do you want to fix makenek.inc?

stgeke:
@RonRahaman Yes please send me your fix

RonRahaman:
@stgeke, I sent you the fixes to NekTests.py in this PR:
https://github.com/stgeke/Nek5000/pull/2

RonRahaman:
Looks like only the eddy and kov_st_state tests fail now.  They fail with this error:
```
 call userchk

        0  0.0000E+00 Write checkpoint:
       0       0 OPEN: eddy_uv0.f00001
byte_write() :: fopen failure!
byte_write() :: fopen failure!
ERROR: Error writing header in mfo_write_hdr.   ierr=  1 
  
call exitt: dying ...
```

PR #227, opened 2017-02-23
High-pass filter stabilization

negips:
High-pass filter relaxation term stabilization.
Parameter space: PARAM(110), PARAM(111), PARAM(112).
Included definitions in PARDICT.
Now uses scratch arrays to reduce memory footprint.


negips:
So the this seems to be the same build failure that happened in the last pull request. Which is related to undeclared variables in the legacy SIZE file. My understanding was that there is some sort of converter which  converts from the old to the new SIZE file (I'm assuming along with variable definitions). Is that also being done for the short tests?

stgeke:
Thanks!

Can you please add a short_test for this new feature.

The converter seems to work but we need to do more testing using different SIZE flavors. I don't want to break the legacy SIZE support without an automatic (using makenek) and robust conversion.

stgeke:
* Let’s say I want to use a dynamic kai which I compute in my .usr file. Do you support this use case? 
* The deltaT dependence on kai is something we have to think about. What happens if the user runs with a variable time step?
* What happens if you use CVODE to solve for your scalars? This solver uses an adaptive time stepper (independent of Nek).

negips:
- Dynamic kai: routine sets the value of kai from param(111) at each timestep. If you wish to do a dynamic calculation in the usr file you can just assign the new value to param(111) at each timestep and your kai will then be dynamic.

- deltaT dependence/variable time step: In normal usage it's not a problem. In physical terms, the energy drain from the system due to the high pass filter remains the same for a particular kai irrespective of time-step. In numerical terms, the filtering makes the simulation unstable only as kai*deltaT approaches (or is larger than) 1. Typical values of kai*deltaT ~ 0.01 - 0.05. So even in variable time-step simulations, it would be fine unless the time-step varies by an order of magnitude or more.

- I'm not familiar with CVODE, so I'm not certain what might be different there. If it uses the bdf-ext type time-stepper, the above points should apply there as well.

I have not really thought of any particular test. Maybe something can be thought of...

Regarding the old SIZE files, how do you want to proceed then? Since the test will not really pass without adding the declarations. The other way is to remove implicit none from the code but I had understood that is to be avoided....

stgeke:
* The step size of CVODE h is typically: 10*dt_nek < h < 10e-5 * dt_nek. Is this problem?
* Let's comment the `implicit none` for now. We fix it later.
* We definitely need a short_test to validate the implementation and also as a HowTo

negips:
Let me get back to you on those points. 
Specifically for CVODE I'll have to see what it does really before I can comment.

lcarasik:
To ensure we start working on the user manual, can you provide a quick write up for each of the parameters you are using the Rea/param file? Also, do you have a quick description of the high pass filter and the reference that can be added as well?

negips:
@lcarasik I could do a short write up and provide the reference. Do you have a format that I need to adhere to or should I just write up a text file?

lcarasik:
@negips Nothing fancy! Just something a generic user can understand and text file would be fine!

stgeke:
Most important for the UG are the parameters.

On 25 Feb 2017, at 16:40, negips <notifications@github.com <mailto:notifications@github.com> > wrote:

@lcarasik I could do a short write up and provide the reference. Do you have a format that I need to adhere to or should I just write up a text file?

lcarasik:
Having a short write up is needed to. Elsewise users won't know what it is.

negips:
I have one more question. Should I put the example in the NekExamples repo or should I put it within the short_tests folder of this repo?

stgeke:
Please provide a verification and validation test case (short_test) not a NexExample.

On 3 Mar 2017, at 16:49, negips <notifications@github.com <mailto:notifications@github.com> > wrote:

I have one more question. Should I put the example in the NekExamples repo or should I put it within the short_tests folder of this repo?

negips:
Ok.

stgeke:
It would be great to merge this pull request. Any chance to provide a testcase and add a paragraph to the user guide?

stgeke:
@negips Can you please update me on this?

stgeke:
@negips Please update me on this otherwise I'll have to assume that you are no longer interested to merge this PR.

negips:
Sorry. I got very occupied with other things lately. Definitely still interested. I should be able to come back to this in a week.

negips:
Right. So I have the stuff almost ready. I have made some documentation, and I have a small test case that can be used as a short test. How do I go about adding these items? Should I make a folder under short_tests and put my test case along with the pdf for the documentation? Also should I modify the NekTests.py to add the case into the tests?

stgeke:
@adampep Can you please help with the short test. Thanks!

stgeke:
Is this ready for review?

stgeke:
Can you please do the following:

* remove bin/genmap and bin/genbox
* add a short description about the test case to the .usr header
* remove the .rea, .map, .box and .f in the case directory
* use the existing .par key "filtering" to select type (supported values are: no, explicit or hpfrt where no is the default)
* use the same .par keys ("addFilterModes" and "filterWeight") as with explicit filtering (or is it really needed to specify chi instead of weight?)
* add an issue on NekDoc about documenting filtering parameters (for .rea and .par)

Thank you for all your efforts!

negips:
Right so I'll just write out the changes in the last update:

- I've cleaned (hopefully all) the temporary files that were generated. Sorry about this, I thought I had cleaned them all last time. Also removed the .map/.f/.rea/.box files.

- The hpfrt parameters are now shared with explicit filtering. So param(103) is the weight/hpfChi and param(101) is the no of additional modes. Same for parameters in the par file (filterWeight and addFiltermodes).

- A new parameter was necessary which acts as a flag for which stabilization procedure is active. This is mapped to param(104). So param(104)=0, 1 or 2 for no stabilization, explicit filtering or hpfrt respectively. For the .par files this is mapped to the values of filtering=no, explicit or hpfrt as was suggested.

- In drive1.f and induct.f the call to perform explicit filtering was based on the filter weight (param(103)) not being 0. Now the checks include param(104)==1 (explicit filter).

- I'll make an issue in NekDoc.

There is something slightly off-topic that I noticed that I'll just mention. In the standard solver (drive1.f) the explicit filtering is performed after the pressure solve (call incomprn).  While in induct.f, which I assume is for the MHD solver, the explicit filtering is performed before the pressure solver. I'm not sure if its intentional, but since they seemed a bit inconsistent, I thought I'll point it out.

stgeke:
Replaced by #389

Issue #115, opened 2016-10-27
modification of uzawa_gmres()


mattiasbr:
Currently, the GMRES algorithm for the pressure in Pn-Pn-2 contains multiplications with operators "U^(-1)" and "L^(-1)" defined as u(i)=sqrt(b(i)) and l(i)=sqrt(binv(i)) in uzawa_gmres_split0(). The reason to have these arrays is to my understanding to have the right inner product, i.e. r^T r = f^T B f (where f is the right-hand side, r = L^(-1) res,  res=B f).

A simpler solution could be to modify the implementation and remove these arrays, by which the inner products would be carried out as
gamma(1) = sqrt(glsc3(r,r,bm2inv,ntot2))
instead of
gamma(1) = sqrt(glsc2(r,r,ntot2)).

(this possible modification was originally suggested to me by James Lottes)

fischer1:
Thanks Mattias - can you issue a pull request for this?

mattiasbr:
Yes, I can do that.

stgeke:
@mattiasbr Can you please submit a pull request or should I close this issue for now.

mattiasbr:
Sorry for being inactive, I will submit a pull request this week.

mattiasbr:
The discussed modifications have been implemented. Currently, these only apply to uzawa_gmres() and not to hmh_gmres() used in PnPn, although similar changes probably could be made to that case as well.

In addition, LGMRES (Baker et al. 2005) has been implemented for Pn-Pn-2. The code has been tested on the 2D-flow around a cylinder (Re=100) and for this simple problem LGMRES yields no improvement over GMRES. However, for more complex problems where the iteration count can be high, LGMRES might yield a significant improvement. Therefore, this solver-option is left in the code but is inactivated.

PR #229, opened 2017-02-28
Issue115

mattiasbr:
Summary of changes:
* uzawa_gmres() is modified so that the inner products are directly weighted with bm2inv, instead of using the arrays ml and mu.
* To detect possible issues in orthogonality of the Krylov vectors due to the classical Gram-Schmidt orthogonalization, modified Gram-Schmidt is turned on after the first restart.
* For high iteration counts where the solver needs to be restarted multiple times, the restart vector might become nearly parallel to the restart vector two restarts ago (due to lack of orthogonality to previous Krylov spaces). To overcome such issues LGMRES (Baker et al. 2005) is implemented (not activated by default). A test has also been implemented for checking the angles between three consecutive residual vectors (commented by default) so the user can assess whether activating LGMRES would be beneficial.

### DIFF
--- a/core/gmres.f
+++ b/core/gmres.f
@@ -10,6 +10,7 @@ c     intype = -1  (implicit)
       include 'SIZE'
       include 'TOTAL'
       include 'GMRES'
+      
       common  /ctolpr/ divex
       common  /cprint/ ifprint
       logical          ifprint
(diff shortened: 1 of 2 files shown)


stgeke:
#115

stgeke:
Can you comment on the improvements you see?

stgeke:
Looks like this PR breaks PnPn

mattiasbr:
The arrays mu_gmres and ml_gmres used by hmh_gmres() were accidentally deleted from GMRES.

stgeke:
I appreciate your effort but we (Paul and me) have decided not to merge it. Feel free to reopen this pull request if there is a strong use case you want to share with us.

PR #247, opened 2017-04-21
Replaced imesh with imsh in the call to setfast in cggo

adampep:


RonRahaman:
I have confirmed that this fixes the Eddy_PsiOmega example, as mentioned in Nek5000/NekExamples#54.  Thanks alot, @adampep!

PR #262, opened 2017-04-27
small corrections to comp_vort3

adampep:


stgeke:
I suggest to use a different name for the real variable. We all think of `j` as an integer.

fischer1:
In this case, we could get rid of the temporary variable j altogether and just
use the jacmi() array (inverse Jacobian).

Sometimes I use "j" if it assists in getting multiples copies of a statement to be one-liners
(thus making the code more readable).

Thanks for finding/fixing this Adam.

fischer1:
Oops... I meant to just comment... not close and comment

Issue #267, opened 2017-04-29
makenek seems to not notice SIZE dependency anymore ?


fischer1:
Hi,

When I edit SIZE, then

makenek blah

should recompile all dependent .f files, but it doesn't do this anymore.

RonRahaman:
I think I see what this issue is.  In makefile.template, the rule for rebuilding the Fortran objects when SIZE changes is:
``` makefile
 $(NOBJS_F) : SIZE
```
Prior to PR #258, the variable `$(NOBJ_F)` was set to be all the Fortran objects.  However, after PR #258, `$(NOBJ_F)` is no longer set to anything.  I'll revert that particular change right now.

RonRahaman:
Fixed by @stgeke

PR #258, opened 2017-04-26
WIP: Bring in latest changes 

stgeke:

PR #268, opened 2017-04-30
Fixes dependency of Fortran objects on SIZE

RonRahaman:
Fixes issue #267 

The Fortran objects should be automatically rebuilt when SIZE changes, but this dependency tracking was broken in PR #258.  

In makefile.template, the rule for rebuilding the Fortran objects when SIZE changes is:
``` makefile
 $(NOBJS_F) : SIZE
```
Prior to PR #258, the variable `$(NOBJ_F)` was set to be all the Fortran objects.  However, after PR #258, `$(NOBJ_F)` was no longer set to anything.

### DIFF
--- a/core/makefile.template
+++ b/core/makefile.template
@@ -88,7 +88,10 @@ TMP1 += $(CMTCORE)
 OPT_INCDIR += -I $(SCMT)
 endif
 
-NOBJS = $(patsubst %,$(OBJDIR)/%,$(TMP1))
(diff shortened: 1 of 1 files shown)


stgeke:
Missed this PR.
Fixed it in https://github.com/Nek5000/Nek5000/commit/fbaac93c9227e7c1743295f2bf57a36c7e922e1c

RonRahaman:
Do we need to recompile the C code if SIZE changes?

PR #276, opened 2017-05-12
correction in axhmsf to get Pn-Pn-2 Fs2_ tests to work

adampep:
This commit reverts partially action of e7da44a commit

### DIFF
--- a/core/subs1.f
+++ b/core/subs1.f
@@ -1418,17 +1418,25 @@ C
       nel   = nelfld(ifield)
       ntot1 = nx1*ny1*nz1*nel
 
-      if (ifaxis.and.ifsplit) call exitti(
-     $'Axisymmetric stress w/PnPn not yet supported.$',istep)
+C      if (ifaxis.and.ifsplit) call exitti(
+C     $'Axisymmetric stress w/PnPn not yet supported.$',istep)
+C
+Cc     icase = 1 --- axsf_fast (no axisymmetry)
+Cc     icase = 2 --- stress formulation and supports axisymmetry
+Cc     icase = 3 --- 3 separate axhelm calls
+C
+C      icase = 1                ! Fast mode for stress
+C      if (ifaxis)      icase=2 ! Slow for stress, but supports axisymmetry
+C      if (matmod.lt.0) icase=2 ! Elasticity case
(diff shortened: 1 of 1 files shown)


stgeke:
@fischer1 Should we revert the changes or is there another solution to this?

fischer1:
I'll try to look asap...

fischer1:
I'm trying to resolve this... The only think I see is below.  Obviously this isn't the difference you're talking about.  I confess I cannot figure out how to look at the two versions in question.

subs1_cur.f subs1_old.f
1421,1422c1421,1431
<       if (ifaxis.and.ifsplit) call exitti(
<      $'Axisymmetric stress w/PnPn not yet supported.$',istep)
---
< c     icase = 1 --- axsf_fast (no axisymmetry)
< c     icase = 2 --- stress formulation and supports axisymmetry
< c     icase = 3 --- 3 separate axhelm calls
<
<       icase = 1                ! Fast mode for stress
<       if (ifaxis)      icase=2 ! Slow for stress, but supports axisymmetry
<       if (matmod.lt.0) icase=2 ! Elasticity case
<       if (.not.ifstrs) icase=3 ! Block-diagonal Axhelm
---
>       if (ifaxis) then
>          icase=2
>       else if(ifsplit.or.matmod.ne.0) then
>          icase=3
>       else if (matmod.eq.0) then
>          icase = 1
>       endif

stgeke:
Here is the diff: https://github.com/Nek5000/Nek5000/pull/276/files

adampep:
To clarify what this pull request does. Commit e7da44a changes the way axhelm and axsf_fast are called in axhmsf. However, neither axhelm nor axsf_fast were changed. The same about all the places axhmsf is called. So or the code was wrong from the beginning and we need new reference values for FS_... tests, or the commit e7da44a was somehow incomplete and this change should be reverted.

fischer1:
I think we should have a skype call on this... I've never been able to figure out what is going on here.

adampep:
OK. Is there a meeting this Friday?

stgeke:
@fischer1 What implementation is correct? Old (this PR) or new?

stgeke:
@fischer1 reviewed the implementation and pushed his change directly to the repo.
@adampep Can you please adjust the target values in the examples. Thanks!

PR #279, opened 2017-05-16
Exact pressure solution

kent0:
Added a exactp subroutine that calculates the exact pressure for the eddy case. In userchk, the pressure error gets written with the 'P err' string identifier. I'm not sure whether I should prescribe the exact pressure in useric, so please give me some guidance on that matter.

### DIFF
--- a/short_tests/NekTests.py
+++ b/short_tests/NekTests.py
@@ -202,6 +202,9 @@ class Eddy_EddyUv(NekTestCase):
         yerr = self.get_value_from_log('Y err', column=-6, row=-1)
         self.assertAlmostEqualDelayed(yerr, target_val=6.489061E-07, delta=1E-08, label='Y err')
 
+        perr = self.get_value_from_log('P err', column=-5, row=-1)
+        self.assertAlmostEqualDelayed(perr, target_val=1.448024E-05, delta=1E-06, label='P err')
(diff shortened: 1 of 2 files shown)


kent0:
I ran the eddy case with nekb as a check, but I now realize that I should've ran the short_tests as well. I'll figure out the issue and push the changes.

RonRahaman:
@kentokaneko, do you have Travis linked to your fork of Nek5000?  

Travis can automatically run the same regression tests on your fork, so you don't get any nasty surprises when you make a PR.  To set it up, follow steps 1 and 2 [here](https://docs.travis-ci.com/user/getting-started/#To-get-started-with-Travis-CI%3A).

kent0:
@RonRahaman No, I don't have that setup yet. Thanks for the tip

stgeke:
What is the motivation for this pull request?

fischer1:
Since we have an exact pressure solution as well as velocity, we should use it --- it's helpful in comparing methods in the future.  Some have better pressure approximations than others.

kent0:
Currently, the pressure error gets written out, but it's not being checked in the test. I will add that functionality

kent0:
The pressure tests have been added, I believe this branch is good to go.

PR #289, opened 2017-05-26
Bug correction raleted to mesh 3 size

adampep:
1) Important for Pipe_Helix
2) corrects drag value in ExtCyl, so delta can be smaller
3) but increases wmax in Hemi

### DIFF
--- a/core/SIZE.inc
+++ b/core/SIZE.inc
@@ -71,9 +71,9 @@ c - - SIZE internals
       parameter (ldimt3=ldimt+3)
 
       integer lx3,ly3,lz3
-      parameter (lx3=lx2)
-      parameter (ly3=ly2)
-      parameter (lz3=lz2)
+      parameter (lx3=lx1)
+      parameter (ly3=ly1)
+      parameter (lz3=lz1)
 
       integer lctmp0,lctmp1
       parameter (lctmp0 =2*lx1*ly1*lz1*lelt)


adampep:
I've checked all the examples and Hemi is not an only one that sets lx3=lx2. So the question is should we set lx3 in SIZE.inc (the original one ant this pull request) or we wove it to SIZE and set it by case

adampep:
I've checked all the tests using originally lx3=lx2 and only hemi fails with lx3=lx1. The most complex formula for lx3 have mhd tests
lx3=3*lx1-2*lx2+ldim*(lx2-lx1)

stgeke:
Why is this different to the legacy SIZE.

adampep:
In the original version lx3 was set independently by the user. Right now is set in SIZE.inc to lx2, which for pn-pn-2 gives you geometrical coefficients calculated on lower resolution mesh. I could understand setting lx3>lx1 for some cases, but never used it myself (always lx3=lx1). Setting it to lx2 is somehow strange to me, but some of examples do it. It would be good if @fischer1  could comment on this. Is lx3 useful as independent variable? Why one would set lx3<lx1? A few examples are affected by this choice, so we have to be consistent here.

fischer1:
lx3 < lx1 is referred to as subparametric (as opposed to isoparametric mappings, lx3=lx1).

It is useful for certain PnPn-2 configurations when you want to ensure that div u = 0 is 
properly enforced.  Otherwise you can get aliasing for the inner products involving div u with
test function q and thus fail to control div u.

An example where this shows up is flow through the frustrum of a cone (I forget if that was a 3D
case or axisymmetric).    Lee Ho and I always liked lx3=lx1; Einar preferred lx3=lx2 - thus we 
had a switch to go back and forth.  It's only in this cone case that I've ever seen it matter.

adampep:
OK. Then we need lx3 in SIZE instead of SIZE.inc. I will prepare proper pull request updating SIZE in short test and nek examples. Current pull request sets lx3=lx1 and is not proper.

stgeke:
I think we can set `lx3=lx1` in `SIZE.inc`. It sounds like this setting works for 99.9% of all usecases. In all other cases you can change `SIZE.inc` or use the legacy SIZE.

PR #301, opened 2017-06-01
added plan5

kent0:
Added plan5 (Richardson extrapolation time-stepper) with a test case, 'eddy_rich' (eddy case with moving boundaries), that tests it's validity.


stgeke:
Is this a variant of plan3? In this case I recommend another name e.g. plan3a.

kent0:
@stgeke It's an explicit Richardson scheme:

First, it takes one O(dt) Euler step to get a solution at the next time step.
Next, it takes two O(dt/2) Euler steps to get solution at the next time step.
Finally, it takes a linear combination of the solutions, 2*O(dt/2) - O(dt), to get an O(dt^2) solution.

You can also repeat these steps to get an arbitrary order time-stepper (currently not implemented). So I think it's sufficiently different than plan3.

fischer1:
@stgeke 

This is a completely new timestepper -- three NS substeps are taken and then
recombined to give O(dt^2) accuracy.   It does not require prior information other
than the solution at time t^n-1 to go to time t^n.

This seems to be important for avoid pressure overshoots in certain situations.

stgeke:
It looks like there is no .rea/.par support for this right?

kent0:
Yes; currently, the user needs to specify

`ifrich = .true.`

 in the usr file.

PR #324, opened 2017-06-09
short test for linear direct/adjoint solver

adampep:
This test is based on the linear stability analysis of 2D channel flow
with Re=2000.
It checks the growth rate of the strongest mode for the linear direct/adjoint
simulation


stgeke:
Can you please add the tests to `.travis.yml`.

PR #325, opened 2017-06-12
short test for linear direct/adjoint solver

adampep:
This test is based on the linear stability analysis of 2D channel flow with Re=2000.
It checks the growth rate of the strongest mode for the direct/adjoint linear simulation


adampep:
This test takes too much time. I didn't realize travis is so extremely slow (on my laptop test takes less than a minute). I will redo it.

PR #331, opened 2017-06-15
ifpert and ifbase renamed to iflin and ifNlin

adampep:


fischer1:
I'm not certain this is a good naming convention?

We can run the base code without nonlinear terms - i.e., unsteady Stokes.

Where does that fit in this convention?

Paul

PR #332, opened 2017-06-20
Using Hypre v2.11.1 for AMG setup

nicooff:
Hypre v2.11.2 is producing wrong results during the AMG setup in some cases. Therefore, it is replaced by v2.11.1 (originally used when developing the setup).

### DIFF
--- a/tools/amg_hypre/hypre/install
+++ b/tools/amg_hypre/hypre/install
@@ -1,5 +1,5 @@
 #!/bin/bash
-wget https://github.com/LLNL/hypre/archive/v2.11.2.tar.gz 
+wget https://github.com/LLNL/hypre/archive/v2.11.1.tar.gz 
 tar -zxvf *.tar.gz
 cd hypre*/src
 ./configure --prefix=`pwd`/../.. CC=gcc FC=gfortran --without-MPI


stgeke:
Any idea why 2.11.2 fails? May be the CHANGELOG can help to understand what's different. In general I like the idea to use the most recent version.

stgeke:
I will merge it for now.
If there is an important feature in HYPRE we want to benefit from we have to make it work with the new version.

stgeke:
An easy fix to make it working with 2.11.2 is:
```
     /* Set parameters (See Reference Manual for more parameters) */
     HYPRE_BoomerAMGSetPrintLevel(solver, print_level);
     HYPRE_BoomerAMGSetOldDefault(solver);
```

nicooff:
I do not know what is causing the problem yet.
Thank you for the fix!

stgeke:
I think it is related interpolation scheme. They changed the default (see CHANGELOG).
 
-----Original message-----

nicooff:
Right, thank you!
It looks like the fix produces large smoothing operators though. After rapid testing, it looks like doing

    /* Set parameters (See Reference Manual for more parameters) */                                                     
    HYPRE_BoomerAMGSetPrintLevel(solver, print_level);                                                                                                                                                                                                                                   
    HYPRE_BoomerAMGSetCoarsenType(solver, coars_strat);                                    
    HYPRE_BoomerAMGSetInterpType( solver, 0 );

produces coarser smoothing operators. And this way, the coarsening strategy is the one given by the user as input and not the default one. Finally, we could let the user choose the interpolation method as well.

PR #333, opened 2017-06-20
added 'SAVE' option to prenek menu

kent0:
Added 'SAVE' option that will write the current mesh configuration to the specified output (default to current 'sesion') without exiting prenek


RonRahaman:
Thanks, @kent0 !

PR #339, opened 2017-06-30
Fix interpolation error with Hypre v2.11.2

nicooff:
The default interpolation strategy between Hypre v2.11.1 and v2.11.2 changed and this caused some issues with the AMG setup and solver.

The default strategy for interpolation is now set manually. Furthermore, the user has now the opportunity to choose the interpolation strategy.


stgeke:
I guess we need to change the `install` script to download 2.11.2 instead of 2.11.1?

nicooff:
Hi, 
The script in the latest commit downloads version v2.11.2. But I realize now that there is still the commit changing to v2.11.1, which should be disregarded.

PR #343, opened 2017-07-18
eddy_par example added

zmiksis:
Setup with .par file for eddy example included in short_tests/eddy_par directory. New directory is needed because Nek automatically uses .par file before old .rea file. Two directories allows for two different test cases.


stgeke:
This is not needed as we have other short_tests for .par

PR #353, opened 2017-08-17
Fix issue 352

nicooff:
Changes dimensions of some face arrays (ifqinp, ifeppm, ifmsfc) that could potentially cause some bugs in 2D. Memory is now stored contiguously also in 2D.


stgeke:
@fischer1 Are you ok with this change?

PR #356, opened 2017-08-24
Adding missing PARDICT names to make it consistent with uparam length

adampep:


adampep:
something wrong with communication with travis as all the test seem to be finished

stgeke:
Do we really need 20 user parameters instead of just 10?

adampep:
Depends on the user. I didn't set size of uparam. It was set to 20 and is kept this way (all the initialisation loops) through the code. Only PARDICT was inconsistent setting 10 instead of 20 names.

stgeke:
I think we are ok as long as PARDICT is using a smaller number. For now I don't see the need to expose 20 user parameters.

adampep:
OK. Anyhow size of uparam in INPUT can be misleading. What would be the goal to keep space in uparam that cannot be set in parameter file?

ggiannako:
Since the space is there, I would be happy if there was support for more than 10 uparams by default. Thus I do not have to modify the source (or use another file) every time I need them (more and more often lately in my simulations).

stgeke:
Ok, lets use `luparam` in the code and set it to 20 in `INPUT`.

stgeke:
btw: There are also other ways to define case specific parameters. You could use `#DEFINE` or use ini-reader to read your (user specific) section in the .par file.

adampep:
I do it in my code. It was mostly for consistency reason. Do you want to rename uparam to luparam?

stgeke:
No I just want to use luparam instead of a hardwired size (e.g. 20) in the code. Sure this won't work in PARDICT.

On 25 Aug 2017, at 14:50, Adam Peplinski <notifications@github.com <mailto:notifications@github.com> > wrote:

I do it in my code. It was mostly for consistency reason. Do you want to rename uparam to luparam?

PR #357, opened 2017-08-28
Introducing luparam to define size of user runtime parameter space

adampep:
- added luparam to SIZE.inc and modified accordingly the rest of files
- check of upram names in par_verify
- pardictkey(77) / 'GENERAL:USERPARAM%%' / is kept in PARDICT for user to see the whole list of possible parameters even though it is not used in par_verify and could be removed


stgeke:
Why SIZE.inc and not INPUT as rparam?

stgeke:
Please ignore my comment.

adampep:
I've just noticed legacy size have to be changed as well

PR #358, opened 2017-08-28
Introducing luparam to define size of user runtime parameter space

adampep:
- added luparam to SIZE.inc and modified accordingly the rest of files
- check of upram names in par_verify


stgeke:
This won't work for old (legacy) SIZE file.
The `luparam` definition must be added automatically to SIZE (by makenek) .

adampep:
OK. I will add this

PR #371, opened 2017-09-21
gen_rea2 routine to write top and bottom parts of newrea file as well

kmittal2:
This new routine gen_rea2 has the same functionality as gen_rea in postpro.f. It also reads the user's rea file to include the top and bottom portion of the rea file which the current routine does not do.


stgeke:
Why not replacing the gen_rea. Do we need two versions?

On 21 Sep 2017, at 19:27, k10 <notifications@github.com <mailto:notifications@github.com> > wrote:

This new routine gen_rea2 has the same functionality as gen_rea in postpro.f. It also reads the user's rea file to include the top and bottom portion of the rea file which the current routine does not do.

  https://github.com/Nek5000/Nek5000/pull/371


Commit Summary 
*	genrea2 routine which also writes top and bottom part of the rea 
*	Merge branch 'master' of https://github.com/Nek5000/Nek5000 into genreamod 
*	changed variable name

File Changes 
*	 M core/postpro.f (244) 

Patch Links: 
*	https://github.com/Nek5000/Nek5000/pull/371.patch 
*	https://github.com/Nek5000/Nek5000/pull/371.diff

ggiannako:
Assuming prenek can read binary files, why not export to re2 directly?

fischer1:
I think it can do that already. (gen_rea)

The point is that a lot of times you want an ascii file.

Ketan was simply trying to streamline the current process where the user has to provide the top and bottom parts of the .rea file, so we could advertise it. 

Here are the issues when doing that, which is why I recommend that for now this should supplement the current process (which works) rather than replace it:

Should gen_rea write out the original .rea file, as it was read in, or should the top part reflect flags that the user might have changed?   If the latter, then we still need the ascii part of the .rea (or .par) file, because we need to have whatever metadata is in there (e.g., comments).
If the former, we can simply read the ascii part and copy it back out.

Same for the lower part.

In my view, there are 100 ways in which construction of a new .rea file could go wrong --- until those are sorted out (and it is difficult), having a working solution as a backup is a good idea.   It seems like this feature is important enough that we should take this step, but it's important to keep the other option until this new feature moves past beta.   I don't think it's too early to include it now, because it's not on the critical path (because there is a working
alternate).

stgeke:
Ok but let’s rename it because my first thought was that gen_rea2 produces a .re2.

On 22 Sep 2017, at 14:01, fischer1 <notifications@github.com <mailto:notifications@github.com> > wrote:

I think it can do that already. (gen_rea)

The point is that a lot of times you want an ascii file.

Ketan was simply trying to streamline the current process where the user has to provide the top and bottom parts of the .rea file, so we could advertise it.

Here are the issues when doing that, which is why I recommend that for now this should supplement the current process (which works) rather than replace it:

Should gen_rea write out the original .rea file, as it was read in, or should the top part reflect flags that the user might have changed? If the latter, then we still need the ascii part of the .rea (or .par) file, because we need to have whatever metadata is in there (e.g., comments).
If the former, we can simply read the ascii part and copy it back out.

Same for the lower part.

In my view, there are 100 ways in which construction of a new .rea file could go wrong --- until those are sorted out (and it is difficult), having a working solution as a backup is a good idea. It seems like this feature is important enough that we should take this step, but it's important to keep the other option until this new feature moves past beta. I don't think it's too early to include it now, because it's not on the critical path (because there is a working
alternate).

fischer1:
Good call. 

Ketan can you come up with a name?  [ BTW, Am I correct that there is a version of this that generates an .re2 file? ]

kmittal2:
We do have gen_re2 currently which does the same thing as gen_rea, but in binary format. I could add a gen_re2_full to complement my gen_rea_full.

PR #373, opened 2017-09-22
Fixes for gfortran and Travis tests

RonRahaman:
These are fixes for problems exposed by Travis CPU tests.  

As of e009caa34110624e5c496a3cc00d1be630bfecd1 , all tests were failing because of (somewhat) pedantic compile-time errors.  The language syntax was allowed by pgfortran (likely the compiler for  local OpenACC tests) but not by gfortran (the compiler for Travis CPU tests).  See further comments below.

Another issue in Travis was simply some debugging print statements.  These extended the length of stdout to the point that Travis was not able to handle the length of the resultant logfile.  This causes tests to fail.  

These allow the Travis tests to compile and run to completion.  A few still fail because of runtime errors, but this PR should help expose the reasons for those failures.

### DIFF
--- a/core/hmholtz.f
+++ b/core/hmholtz.f
@@ -3075,8 +3075,8 @@ C-------------------------------------------------------------------
       include 'MASS'
       include 'EIGEN'
 
-c     common  /cprint/ ifprint
-c     logical          ifprint
+      common  /cprint/ ifprint
+      logical          ifprint
 
       real w1   (lx1,ly1,lz1,lelt),
      $     w2   (lx1,ly1,lz1,lelt)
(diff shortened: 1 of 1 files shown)


core/hmholtz.f
```diff
@@ -3075,8 +3075,8 @@ subroutine chktcg1_acc(tol,res,h1,h2,mask,mult,imesh,isd)
       include 'MASS'
       include 'EIGEN'
 
-c     common  /cprint/ ifprint
-c     logical          ifprint
+      common  /cprint/ ifprint
+      logical          ifprint
```
RonRahaman, line 3079: This fixes a type-casting error on line 3170.

core/hmholtz.f
```diff
@@ -3364,9 +3364,9 @@ subroutine axhelm_acc_debug (au,u,helm1,helm2,imesh,isd)
      $ ,             tmp2  (lx1*ly1*lz1*lelt)
      $ ,             tmp3  (lx1*ly1*lz1*lelt)
 
-      real           tm1   (lx1*ly1*lz1,lelt)
-      real           tm2   (lx1*ly1*lz1,lelt)
-      real           tm3   (lx1*ly1*lz1,lelt)
+      real           tm1   (lx1*ly1*lz1*lelt)
```
RonRahaman, line 3367: I believe this is the intended dimensionality.  The previous declaration:
`real tm1 (lx1*ly1*lz1, lelt)`
failed at line 3525.

PR #379, opened 2017-09-28
Modify .f file header to add pressure mesh flag

adampep:
Currently pressure for p_n-p_n-2 can be written in .f files on velocity or pressure meshes depending on if_full_press flag. Unfortunately this is not consistent between writing and reading routines as writing does not assume anything (up to user) and reading assumes single and double precision to have pressure on velocity and pressure meshes respectively. Adding flag to the header could clarify this issue. In the future it could be used as well in visualization software to properly plot pressure from restart files.

### DIFF
--- a/core/ic.f
+++ b/core/ic.f
@@ -2329,17 +2329,28 @@ c-----------------------------------------------------------------------
       include 'SOLN'
       include 'PARALLEL'
       include 'RESTART'
+      include 'TSTEP'
 
       character*132 hdr
       character*4 dummy
+      logical if_press_mesh
 
       p0thr = -1
+      if_press_mesh = .false.
 
       read(hdr,*,iostat=ierr) dummy
      $         ,  wdsizr,nxr,nyr,nzr,nelr,nelgr,timer,istpr
      $         ,  ifiler,nfiler
      $         ,  rdcode      ! 74+20=94
+     $         ,  p0thr, if_press_mesh
+
+      if (ierr.gt.0) then ! try again without pressure format flag
+        read(hdr,*,iostat=ierr) dummy
+     $         ,  wdsizr,nxr,nyr,nzr,nelr,nelgr,timer,istpr
+     $         ,  ifiler,nfiler
+     $         ,  rdcode      ! 74+20=94
      $         ,  p0thr
+      endif
 
       if (ierr.gt.0) then ! try again without mean pressure
         read(hdr,*,err=99) dummy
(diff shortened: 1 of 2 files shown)


adampep:
I've just tested replacing P with Q and it doesn't work. visit does not recognize pressure index and does not give possibility to plot it. Version with additional flag works fine.

fischer1:
I guess that, no matter, what, there is no possibility of VisIt plotting the pressure that's on the gl points, correct?

stgeke:
Right now, that's correct. 

-----Original message-----

adampep:
No. I wasn't clear in my last post. visit reads header, stops at pressure and doesn't read properly mesh, temperature or passive scalars. There is not way to plot anything what is located after pressure, but even velocity plot gets completely wrong.

fischer1:
For my part, I can't see any downside to changing the header per Adam's suggestion?
That's why we have, say, "std4" to indicate 4-byte reals in the file header.

We could have, say, pn28, to indicate pressure on n-2, 8 byte reals?

To my knowledge, we are not using this discriminator yet (and also not using it in VisIt),
so it should be ok to proceed in this way??

stgeke:
Do you suggest to put pn28 in the excoder?

On 3 Oct 2017, at 17:50, fischer1 <notifications@github.com <mailto:notifications@github.com> > wrote:

For my part, I can't see any downside to changing the header per Adam's suggestion?
That's why we have, say, "std4" to indicate 4-byte reals in the file header.

We could have, say, pn28, to indicate pressure on n-2, 8 byte reals?

To my knowledge, we are not using this discriminator yet (and also not using it in VisIt),
so it should be ok to proceed in this way??

fischer1:
Right now, the header looks like:

#std 4  8  8  1         96         96  0.2000000000000E-01       200      0      1 XUP         0.0000000E+00 

The intent was #std 4 could later change to (say)  #pn2 8 .... or whatever.
"std" was the first tag we used... with the intent to later support non-standard file formats.

As far as I know, we don't currently trigger off of the "std" keyword, nor does VisIt.

As far as I know, changing "std" to pn2 would have no negative consequences and would then allow us to trap for the pn2 keyword to know that we need to treat the incoming pressure with a copy rather than interpolation ?

adampep:
Just checked. Replacing #std with #pn2 does not work. Visit seem to check it

stgeke:
I think Adam's solution adding a logical flag (after the thermodynamic pressure) is fine.
@fischer1 What do you think?

PR #384, opened 2017-10-03
ortho res in esolve to make sure it is orthogonal to null space

kmittal2:
A fix to ensure that the residual is orthogonal to the null space when solving for pressure.
cc: @fischer1


kmittal2:
@stgeke I am seeing some strange error messages in the log for Travis' checks (https://travis-ci.org/Nek5000/Nek5000/jobs/282914377)

PR #387, opened 2017-10-05
Remove change to right hand side during scalar solve at the end of ngeom = 2

kmittal2:
This PR removes a call to a routine which was changing the rhs for the scalar solve.. this affected calculations with ngeom > 2.
cc: @fischer1

### DIFF
--- a/core/conduct.f
+++ b/core/conduct.f
@@ -106,7 +106,6 @@ C        Radiation case, smooth convergence, avoid flip-flop (ER).
 
  1000    continue
  2000    continue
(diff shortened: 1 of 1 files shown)


emerzari:
Thanks Ketan, I have also tested on my side and this change does the trick.

kmittal2:
Thanks Elia, glad we caught this!

stgeke:
Why didn’t we catch that with our regression tests?

emerzari:
There is no test that exercises ngeom>2 in my understanding.

PR #399, opened 2017-10-23
Ensemble mode

makavakh:
These additions allow running in Ensemble mode, i.e. running independent sessions in parallel with a single run by invoking ENSEMBLE flag as a PPLIST option in makenek. Requires SIZE file with multimesh parameters (corresponding to the latest multimesh parameters) and a submission script 'nekens' added to the bin directory.

Can be run like this: ./nekens casename number-of-sessions np-per-session
Ex: ./nekens Channel 4 8


stgeke:
It would be great if this PR would include a generic routine to average a given field across all sessions.

makavakh:
Yes, I'll add that shortly.

stgeke:
@makavakh Can you please push your latest changes. Thanks!

makavakh:
@stgeke Just added my latest changes.

kmittal2:
Quick question: why do we have nel,lx,ly,lz as input to routine msgsop_get_hndl(hndl,nel,lx,ly,lz).
It could probably just be msgsop_get_hndl(hndl,nel) since lx,ly and lz are already known. Even nel is known but 'nel' gives user to input nelv or nelt if needed.

stgeke:
I think we need nel,nx,ny,nz to support all combinations of Pn/Pn, Pn/Pn2, v-mesh, t-mesh.

-----Original message-----

kmittal2:
got it.. in that case we could have just nx instead of nx,ny,nz?

stgeke:
Replaced by #481

PR #409, opened 2017-11-13
make postnek compatible with .par + .re2

kent0:


kent0:
I will update this PR with the par parser so that postnek can support re2 + par.

stgeke:
Can you please include prenek in this PR?

kent0:
Prenek currently does not accept rea+re2 or par+re2. If we alter it to be able to read re2, we need to add an option at the beginning for the user to select which format they want: rea, rea+re2, par+re2. We would need to change how the code reads/writes. Also we need to change aspects of the program that depends on knowing that the output is an rea file ie., this might take longer.

I'll keep you posted on the progress. I plan on getting the postnek changes completed soon.

stgeke:
Ok, in this case let's split it in two.

stgeke:
Can you please update me on this.

stgeke:
Again, please update me on this otherwise I have to close this PR.

kent0:
The code I have right now compiles with the par reader routines borrowed from Nek5000. I'm still debugging issues that arise from par files (Old functionality of reading rea files seems to work fine). I hope to finish this up in the next couple of days.

kent0:
I've verified that the new postnek works by reading the output of conj_ht case (.par + .re2) with cvode turned off (I couldn't get cvode to work for some reason, but that's a separate issue I'll investigate). I think it's good to go.

PR #411, opened 2017-11-14
planar averaging with gs_op

kmittal2:
Routine for planar averaging with gs_op in x/y/z direction for box type meshes.

### DIFF
--- a/core/navier5.f
+++ b/core/navier5.f
@@ -4691,48 +4691,84 @@ c     enddo
       return
       end
 c-----------------------------------------------------------------------
-      subroutine plan_avg_tensr(ua,u,gs_hndl,nelx,nely,nelz,ifld,idir)
+      subroutine plan_tensr_op(ua,u,gs_hndl,nelx,nely,nelz,ifld,idx,op)
       include 'SIZE'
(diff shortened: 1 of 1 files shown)


stgeke:
Will this work for pressure and PnPn2? 
How about making it more generic i.e. support other gs operations like min/max?

kmittal2:
For pressure in PnPn2, I think it will be better if the user maps the pr to pm1 and then send it to this routine. Or we could add another input to the routine which determines if the input variable needs to be mapped to mesh1 or not.
I will add support for other gs operations (min and max should suffice I think).

stgeke:
I'll wait for #399 to be ready. Both should provide a similar API.

kmittal2:
I don't think we really need ifld here.. my intention for ifld was to allow user to let the user average across just fluid elements, or across fluid and solid elements.. however I just realized that if we consider that we number fluid elements first, for cases where nelt > nelv, the global numbering used to setup gs_hndl won't be consistent to allow users to average a variable across fluid and solid elements.

stgeke:
Replaced by #481

PR #420, opened 2017-11-17
Required by mhd tests

adampep:


stgeke:
Fixed in https://github.com/Nek5000/Nek5000/commit/fa72bcc5336be9768f2d2b7601f5c011d998c585

PR #428, opened 2017-11-21
Added gslib documentation

nicooff:
Hi,

Two years ago, I wrote this documentation for the gs library. I haven't modified it since then and I believe some things could be added. However, I leave it here for the moment as I think it provides a good basis for understanding how gslib works from a practical point of view.

Nicolas


stgeke:
This is not the gslib repo. Please submit this PR to gslib/gslib

PR #432, opened 2017-11-27
Correction to reduce number of pressure iterations for cray compiler

adampep:


stgeke:
Did you change anything else? I think the current Cray compiler FFLAGS are wrong. Instead of `-s default64` we should use `-s real64` otherwise integer is also 8bytes.

stgeke:
Included in https://github.com/Nek5000/Nek5000/commit/6980e4cf9cf0ad47b0241333ea5dcf8c8ef9a805

PR #423, opened 2017-11-20
OpenACC memory optimizations

RonRahaman:
This PR introduces several memory optimizations in the OpenACC kernels for axhelm and hsmg_do_fast GPU kernels.  Generally, we have done the follwoing.  
* Shared memory buffers are used strategically.  The buffers are declared with the ACC `gang private` and `cache` directives.
* Array accesses are reordered to minimize memory transactions per request.

Using the singlerod example on both pascal and volta architectures, we observe 2x and 3x increase in FLOP/s for the ax and hsmg_do_fast kernels, respectively.

PR #436, opened 2017-11-30
OpenACC setprec_acc on device

RonRahaman:
This implements `setprec_acc` on device.  It allows us to remove the host/device updates surrounding the call to the previous, host-only implmentation of `setprec_acc`.


misunmin:
This version fails too before going into the first pressure solve.


We can now do testings/correctness checking on summitdev/titan before PR.


________________________________

misunmin:
Same failure on local machine Tesla.


________________________________

emerzari:
Should we revert the commits?

________________________________

misunmin:
Yes

________________________________

RonRahaman:
Okay, I created a PR to revert this one.  I had a version working on tesla, but I must've messed up it up when I merged upstream changes.  

@misunmin, could you let me know how you're doing testing on summitdev and titan?  I'm not really in the loop about that.

misunmin:
We need to revert #423 too. Will reply there.

________________________________

PR #440, opened 2017-11-30
Update README.md

rajeeja:
Instructions to build exo2nek


stgeke:
I don't think this needed. All the tools have to be compiled using our maketools script.

PR #468, opened 2018-01-25
Fix projection always being turned on for velocity

nicooff:
The modification fixes the fact that velocity projection is always being turned on when using PnPn-2. Projection is turned on only if variable `residualProj=yes` in .par file, irrespectively of the value of param(94).


nicooff:
I close it since the issue seems to be different.

PR #472, opened 2018-02-22
Bvflow par

dshaver14:
Added par file support for fixed flow rate


emerzari:
Adding .par support for this is a very good idea. 
Let us be cautious in supporting variable properties which will need to be periodic as well for the problem to be well posed.  It is not really about the fixed volumetric flow rate as much as periodic boundary conditions. 
 My suggestion is that, if the user selects both this option and variable properties a warning be issued suggesting to ensure properties are periodic.

stgeke:
According to this PR one needs to set

[PROBLEMTYPE]
fixedFlow = velocityX
flowRate = 1.0

in .par to fix the flow rate such that the bulk velocity (in x) is equal to 1.
I find this difficult to grasp from the parameter names above. 

Any suggestions how to improve this?

dshaver14:
Keep in mind it also supports

[PROBLEMTYPE]
fixedFlow = volumetricX 
flowRate = 1.0

which will force the volume flow rate to 1.0.  I am certainly open to suggestions to changing the syntax,  I was trying to balance between legibility and keeping the number of parameters to a minimum.

stgeke:
Is this prescribing the volumetric or the mass flux?

dshaver14:
It's described as volumetric in drive2, but I've only ever used it with fixed velocity.

stgeke:
I don't think the section PROBLEMTYPE is right. These parameters belong to a periodic boundary condition. I'll wait for #475 before merging this PR.

lcarasik:
Can we please implement this soon? 

It's been almost 3 years now and it would be useful to have this properly implemented in .par for Nek5000.

PR #476, opened 2018-03-10
allow comp_vort3 to compute the curl of an arbitrary vector for cht cases

kent0:
By setting ifield to 2 and using bintm1 which has nonzero values inside the solid domain, a user can compute the curl of an arbitrary vector in the case of cht cases.

### DIFF
--- a/core/navier5.f
+++ b/core/navier5.f
@@ -852,18 +852,28 @@ c        vvy=vr(i)*rym1(i,1,1,e)+vs(i)*sym1(i,1,1,e)+vt(i)*tym1(i,1,1,e)
 c
 c    Avg at bndry
 c
+
       ifielt = ifield
       ifield = 1
+      if (ifheat) ifield = 2
       if (if3d) then
          do idim=1,ldim
             call col2  (vort(1,idim),bm1,ntot)
             call dssum (vort(1,idim),lx1,ly1,lz1)
-            call col2  (vort(1,idim),binvm1,ntot)
+            if (ifheat) then
+               call col2  (vort(1,idim),bintm1,ntot)
+            else
+               call col2  (vort(1,idim),binvm1,ntot)
+            endif
          enddo
       else
          call col2  (vort,bm1,ntot)
          call dssum (vort,lx1,ly1,lz1)
-         call col2  (vort,binvm1,ntot)
+         if (ifheat) then
+            call col2  (vort,bintm1,ntot)
+         else
+            call col2  (vort,binvm1,ntot)
+         endif
       endif
       ifield = ifielt
 c


stgeke:
How about passing iftfld. This way it’s more generic but not backward compatible as the API is different.

On 10 Mar 2018, at 13:25, Kento Kaneko <notifications@github.com <mailto:notifications@github.com> > wrote:

By setting ifield to 2 and using bintm1 which has nonzero values inside the solid domain, a user can compute the curl of an arbitrary vector in the case of cht cases.

  https://github.com/Nek5000/Nek5000/pull/476


Commit Summary 
*	use ifield = 2 and bintm1 when ifheat is true

File Changes 
*	 M core/navier5.f (14) 

Patch Links: 
*	https://github.com/Nek5000/Nek5000/pull/476.patch 
*	https://github.com/Nek5000/Nek5000/pull/476.diff

fischer1:
Here is the main issue as I see it:

1) There is no difficulty with computing vorticity everywhere.  If it's a velocity field, it will be zero in the solid elements.

2) The difficulty arises from the "dsavg" .

       Here, if it's a velocity field, then the user probably wants a jump in vorticity at the fluid-solid interface, meaning zero in the solid and nonzero in the fluid.    This jump can be realized if we do the dssum with ifield=1 . (gs_lib is called with a different handle).

If the user wants curl of a vector field that includes solid and fluid elements, then we have to set ifield=2 in the case of CHT.

Another issue is the use of bm1/binvm1 or bm1/bintm1 in the dsavg process.

One issue with adding another flag is that we already have a flag --- ifield --- that does
the same thing.     Here's an idea - can we simply have a routine comp_curl_all() ?

(In this case, "curl" seems a more appropriate name that vorticity.)

Or - rather than a "solid/fluid" flag, perhaps an "ifavg" flag.    I was hoping there was
such a flag already, because the user could then simply set it to false and be done.

stgeke:
I agree with I just want to avoid the ifield magic behind the scenes (like in other subroutines). I prefer to make it more transparent by passing it somehow as an argument. Same with e.g. dssum().

but I don’t like the hidden 

On 10 Mar 2018, at 14:51, fischer1 <notifications@github.com <mailto:notifications@github.com> > wrote:

Here is the main issue as I see it:

1.	

There is no difficulty with computing vorticity everywhere. If it's a velocity field, it will be zero in the solid elements.

2.	

The difficulty arises from the "dsavg" .


Here, if it's a velocity field, then the user probably wants a jump in vorticity at the fluid-solid interface, meaning zero in the solid and nonzero in the fluid.    This jump can be realized if we do the dssum with ifield=1 . (gs_lib is called with a different handle).


If the user wants curl of a vector field that includes solid and fluid elements, then we have to set ifield=2 in the case of CHT.

Another issue is the use of bm1/binvm1 or bm1/bintm1 in the dsavg process.

One issue with adding another flag is that we already have a flag --- ifield --- that does
the same thing. Here's an idea - can we simply have a routine comp_curl_all() ?

(In this case, "curl" seems a more appropriate name that vorticity.)

Or - rather than a "solid/fluid" flag, perhaps an "ifavg" flag. I was hoping there was
such a flag already, because the user could then simply set it to false and be done.

fischer1:
I agree.  I don't like ifield either.   

How about just a plain "compute_curl" routine --- no averaging added.

That would cover all cases?

We have gradm1, this would be similar.

stgeke:
Yes no averaging make sense but we need to know when to use nelv/nelt.

fischer1:
Why not just compute curl everywhere, all the time?

If velocity is zero in the solid, it will return zero (except for the interface averaging issue).

?

stgeke:
True

On 11 Mar 2018, at 14:19, fischer1 <notifications@github.com <mailto:notifications@github.com> > wrote:

Why not just compute curl everywhere, all the time?

If velocity is zero in the solid, it will return zero (except for the interface averaging issue).

?

kent0:
I've been looking around and found a subroutine called `curl` that is essentially the same thing as `comp_vort3` (except `ntot = lx1*ly1*lz1*nelv`, in induct.f). Also there's another routine called op_curl in plan4 that's implemented differently. It seems like there are a lot of 'curl' routines that are floating around in the code. Should I implement the non-averaging curl routine as `curl` or would that cause backward compatibility issues?

stgeke:
I think all three subroutines
- curl(vort,u,v,w,ifavg,work1,work2)
- comp_vort3(vort,work1,work2,u,v,w)
- op_curl(w1,w2,w3,u1,u2,u3,ifavg,work1,work2)

do more or less the same thing but there are some small differences
- averging (dssum)
- imesh handling
- ifaxis handling

I guess it makes sense to clean up the mess and come up with a single generic version. 
@fischer1 What do you think?

stgeke:
Is anybody working on this?

stgeke:
Looks like this needs a bigger discussion. I'll close this for now.

PR #481, opened 2018-03-29
handle planar averaging and multi-session averaging

kmittal2:
Combining #411 and #399 and adding a wrapper to combine the functionality of these two PRs.


stgeke:
- looks like you commited some files in /bin which are not necessary
- instead of ensemble averging we should think of it as a generic gs_op across sessions 
- I prefer one get_session_info() for both coupled multisession and uncoupled multisession (here ensemble) with no additional PPLIST symbol

stgeke:
Let me add some addtional suggestions:

- Instead of new short_test let's test the multisession gs_op in the existing eddy_neknek

- Let's make the API more consistent - what do you think of the following API:
```
c global tensor product gsop
gtpgsop_get_hndl(hndl,dir,nelx,nely,nelz)
gtpgsop(ua,u,op,hndl)

c multisession gsop
msgsop_get_hndl(hndl,nel,nx,ny,nz)
msgsop(ua,u,op,hndl)
```
- Averaging routines like savg() are not needed as gsop with sum is good enough

kmittal2:
API looks good.. I think it is a good idea to leave averaging routines in there because it is not always clear what the correct weights are (straightforward for multi-session case where weight = 1/nsessions, but not so much for planar averaging which requires collocation with bm1)

stgeke:
Let me do some modifications of your PR. I'll come back to you as soon as the major design issues are resolved.

kmittal2:
sounds good

stgeke:
Implemented in https://github.com/Nek5000/Nek5000/pull/491

PR #485, opened 2018-04-03
Support for HEX20/QUAD8 elements

ggiannako:
This update sets HEX20 and QUAD8 the default element types for exo2nek to make it compliant with the ExodusII standard and in accordance to the current .re2 file capabilities. HEX27 and QUAD9 are still supported to maintain backward compatibility (a warning is issued).

### DIFF
--- a/tools/exo2nek/README.md
+++ b/tools/exo2nek/README.md
@@ -1,11 +1,11 @@
-Reads a exodus II (.exo) file and generates a .re2 file containing the grid,
-curved sides and sideset ids. 
+Reads an exodus II (.exo) file and generates a .re2 file containing the grid,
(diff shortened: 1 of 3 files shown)


stgeke:
@emerzari Can you please test this with an ICEM Exodus file

emerzari:
We should add a short test case for the mesh conversion. This is an important capability and it should be tested.

stgeke:
Good idea. Actually we should test all our tools. Right now we only test genmap. This requires some changes to the testing framework.

On 10 Apr 2018, at 14:08, emerzari <notifications@github.com <mailto:notifications@github.com> > wrote:

We should add a short test case for the mesh conversion. This is an important capability and it should be tested.

PR #519, opened 2018-07-12
Fixed small bugs in amg_hypre.

nicooff:
Fixed 2 bugs in amg_hypre:
- We check that the maximum number of levels is high enough. If not, the setup stops and the user is asked to run again with a higher value.
- No more segmentation fault if amgdmp_*.dat files are not present.

### DIFF
--- a/tools/amg_hypre/amg_hypre.c
+++ b/tools/amg_hypre/amg_hypre.c
@@ -20,7 +20,7 @@
     his Ph. D. thesis "Towards Robust Algebraic Multigrid Methods for 
     Nonsymmetric Problems".
 
-    - Last update: 30 June, 2017
+    - Last update: 12 July, 2018
 */
 
 int main(int argc, char *argv[])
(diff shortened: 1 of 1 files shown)


stgeke:
It looks like user input is ignored and maxlevels is always set to 30. Removing this (unintended) reset I get 

Error: increase maximum number of levels and run again.

But output files work fine?

nicooff:
Actually, max number of levels should not be constrained by the user. Otherwise, the AMG breaks (I had not realized that before). The option is there only if the default value of 30 is too low and the user needs to use a larger value. Which could happen for extremely large meshes (even if never observed). 

Maybe not the best design choice. I'll just remove the option and hard code a high number to be safe. That's probably less confusing.

nicooff:
Additional fixes:
- Max. number of coarsening levels is not a user input any more. A default value of 50 is used instead.
- The smoother tolerance cannot be <0.
- Removed some interpolation options that were causing the setup to crash.

stgeke:
Included in #557

PR #520, opened 2018-07-13
Import latest 

stgeke:


stgeke:
Note, this PR is not ready for merging yet ...

I don't think you need a c++ compiler but we better double-check

nicooff:
I actually need to provide a c++ compiler to build hypre. It would not compile on our local supercomputer without it. Imight be doing something wrong but could you please double check?

stgeke:
The manual says:

hypre is written in C (with the exception of the FEI interface, which is written in C++) and provides an interface for Fortran users.

I think setting CXX=“” builds HYPRE with no C++ support.

On 13 Jul 2018, at 17:10, Nicolas <notifications@github.com <mailto:notifications@github.com> > wrote:

I actually need to provide a c++ compiler to build hypre. It would not compile on our local supercomputer without it. Imight be doing something wrong but could you please double check?

stgeke:
Ok it was just confirmed: It is C only. The C compiler can actually be a C++ one.

nicooff:
Thank you!
I was still using version 2.11.2, which might be the reason.

PR #526, opened 2018-08-20
N2to3 r8

dshaver14:
Added R8 flag to n2to3.


stgeke:
Please set FFLAGS in tools/makefile instead (see e.g. genmap).

PR #528, opened 2018-08-28
Set nmxh to 1000 when setting IC with presolve solution

kent0:
nmxh is set to 100 or 1000 depending on iftran; but for the presolve option, it should be set to 1000 to be consistent with the behavior when iftran is false.


stgeke:
Instead we should implement https://github.com/Nek5000/Nek5000/issues/534

PR #531, opened 2018-09-05
quick fix  to ms_glsc2 declaration in PARALLEL

kmittal2:
just a quick fix in PARALLEL.. don't need to change function declaration in math.f because each function includes PARALLEL through TOTAL.

### DIFF
--- a/core/PARALLEL
+++ b/core/PARALLEL
@@ -48,6 +48,3 @@ c     Maximum number of elements (limited to 2**31/12, at least for now)
 
       integer               nsessions
       common /session_info/ nsessions
-
-      real ms_glmin,ms_glmax,ms_glamin,ms_glamax,ms_glsum,ms_glsc3
-     $     ,ms_glsc2
--- a/core/PARALLEL
+++ b/core/PARALLEL
@@ -50,4 +50,4 @@ c     Maximum number of elements (limited to 2**31/12, at least for now)
       common /session_info/ nsessions
 
       real ms_glmin,ms_glmax,ms_glamin,ms_glamax,ms_glsum,ms_glsc3
-     $     ms_glsc2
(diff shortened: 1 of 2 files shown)


stgeke:
I prefer correct declarations.

On 5 Sep 2018, at 23:06, k10 <notifications@github.com <mailto:notifications@github.com> > wrote:

just a quick fix in PARALLEL.. don't need to change function declaration in math.f because each function includes PARALLEL through TOTAL.

  https://github.com/Nek5000/Nek5000/pull/531


Commit Summary 
*	fix ms_glsc2

File Changes 
*	 M core/PARALLEL (2) 

Patch Links: 
*	https://github.com/Nek5000/Nek5000/pull/531.patch 
*	https://github.com/Nek5000/Nek5000/pull/531.diff

kmittal2:
That seems to require that the user declare ms_glsc2 as real in the subroutine that they want to call it in as well.. I think we should avoid it.
Even if we include this in Parallel, we will have to assume that the user has included Parallel in the subroutine that they are calling ms_glsc2 from. To avoid this, should we consider renaming these functions such that they don't default to integers?

pschlatt1:
I think it is quite dangerous to have in a code where the typical user does not use IMPLICIT NONE non-integers that default to integers. I would definitely recommend renaming (which then of course means that prefixes are not a good solution as they might be used for all types to variables.)

stgeke:
We need to explicitly define all functions anyway to support IMPLICIT NONE in .usr files. For best visibitly I prefer to do this through SIZE. This way we're safe and renaming is not needed.

-----Original message-----

stgeke:
Fixed in https://github.com/Nek5000/Nek5000/commit/597cdbccd56496dfaf3a725e5598b0e4c3406067

PR #537, opened 2018-09-24
New TML bc

dshaver14:
This commit makes changes to the SYM boundary, but I think we were discussing adding an all new BC type.  Really this is a wall BC with a prescribed shear stress.


dshaver14:
The functionality of this boundary condition is already fully supported by the shl BC.

PR #549, opened 2018-11-01
Neknek latest updates

kmittal2:
creating initial commit for new NekNek stuff... this fix is for moving meshes. was missing a call to multimesh_create for pnpn in drive2.f


kmittal2:
Summary of changes made so far:
- bmg,upf for integration across overlapping domains
- rescale_x_ms 
- Mass-flux correction at interface boundaries (subroutine fix_surface_flux called in bcdirvc)
- vol_flow_ms (called at igeom.eq.ngeom when ifneknekc)
- multimesh_create is called at igeom = 3 so that we can solve with extrapolated data at igeom = 2 for moving meshes.
- neknek/neknekb/neknekb_mpmd for arbitrary number of sessions

core/bdry.f
```diff
@@ -698,6 +698,7 @@ SUBROUTINE BCDIRVC(V1,V2,V3,mask1,mask2,mask3)
       CALL ADD2(V2,TMP2,NTOT)
       IF (IF3D) CALL ADD2(V3,TMP3,NTOT)
 
+      if (ifneknek) call fix_surface_flux
```
stgeke: I think we should check ifneknekc instead of ifneknek?

core/drive1.f
```diff
@@ -96,6 +96,8 @@ subroutine nek_init(comm_out)
 
       if (ifmvbd) call setup_mesh_dssum ! Set mesh dssum (needs geom)
 
+      if (ifneknek) call init_neknek_par
```
stgeke: I think we should check ifneknekc instead of ifneknek?

core/multimesh.f
```diff
@@ -12,6 +12,19 @@
...
+
+      ntot1 = lx1*ly1*lz1*nelt
+c     Initialize unity partition function to 1
+      call rone(upf,ntot1)
+      call col3(bm1ms,bm1,upf,ntot1)
+
+      return
+      end
+c-------------------------------------------------------------
       subroutine multimesh_create
```
stgeke: We should rename this subroutine to setup_neknek

core/multimesh.f
```diff
@@ -522,3 +537,680 @@ subroutine neknek_xfer_fld(u,ui)
...
+        if (nid.eq.0)
+     $    write(6,'(i2,i8,i4,1p4e13.4,a11)') idsess,istep,ictr,time,
+     $    dvxmax,dvymax,dvzmax,' del-vol-vxy'
+
+      enddo
+
+      return
+      end
+c-----------------------------------------------------------------------
+      function rs_glmin(a,n)
```
stgeke: What's the difference to the existing glmin_ms()?

short_tests/NekTests.py
```diff
@@ -588,6 +588,149 @@ def tearDown(self):
...
+            ldim='2',
+            lx1='8',
+            lxd='12',
+            lx2='lx1-2',
+            lelg='1000',
+            nsessmax='2',
+        )
+
+        self.build_tools(['genmap'])
+
```
stgeke, line 608: We need to add: 
self.run_genmap(tol='0.01')

short_tests/eddy_neknek_mv/README
```diff
@@ -0,0 +1,92 @@
+c
```
stgeke: Please remove this file. It's not needed for a short_test.

stgeke:
This pull request should also refactor the MPI communicator assignment for different sessions.

core/multimesh.f
kmittal2: These are not needed. I accidentally carried them over from my dev branch.

PR #570, opened 2018-12-14
RANS for V18

dshaver14:
The calculation of k and omega source terms are split from the calculation of eddy viscosity.
Eddy viscosity is now updated as it is called
Source terms are now updated as they are called
Should be independent of whether k or omega is first
Added warnings for low-mach (not yet supported)


dshaver14:
replaced with pr #572

PR #573, opened 2018-12-16
lpm

stgeke:


stgeke:
Two remaining problems:
* odd particle parition sizes on 3 or 4 ranks using lpm_one
* why is the projection error on multiple ranks smaller than on one

PR #574, opened 2018-12-17
Stops compilation on my linux distribution.

adampep:
It seems to stop compilation of all the examples as well.

### DIFF
--- a/3rd_party/cvode/install
+++ b/3rd_party/cvode/install
@@ -1,6 +1,8 @@
 #!/bin/bash
 set -e
 
+VER=2.7.0
+
 if [ "$1" == "clean" ]; then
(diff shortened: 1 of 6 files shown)


stgeke:
Hmm that's odd. Anyway can you please do the proposed change for all install scripts:

./3rd_party/cvode/install
./3rd_party/gslib/install
./3rd_party/hypre/install
./3rd_party/parRSB/install
./tools/amg_setup/hypre/install
./tools/exo2nek/3rd_party/install

adampep:
Updated. There is some problem with downloading tar file for parRSB

stgeke:
That's ok - there is no parRSB release yet.

PR #586, opened 2019-01-14
Import latest changes

stgeke:


hackljf:
0af2427 is a VERY nice touch and very useful. Thank you so much for putting this in!

PR #614, opened 2019-04-09
Fix bug in makeq_aux

ggiannako:
When dp0dt=.true. and we solve for T using Helmholtz, the dp0dt term was not added correctly since nxyz was not defined.


stgeke:
Thanks!

PR #625, opened 2019-05-31
amg for neknek

kmittal2:
This pull request enables amg for neknek where the usr can read in multiple amg files from the same folder. The file names are based on the mesh name by default, and can be redefined in usrdat2.


stgeke:
Replaced by #632

PR #660, opened 2020-05-19
Update add_temp

kent0:
Fix cht issue and improve consistency with last ifield


stgeke:
What was the issue you're trying to fix?

kent0:
After the initial setup for a standard cht case, ifldmhd = npscal + 3 = 3. When I call add_temp(   ,   ,1) and dssum is called when ifield = 3, Nek handles it as a mhd field and applies the operation in fluid domain only.

The other part of the PR is so that the setting of the new field is consistent with the field that is being copied from.

PR #667, opened 2020-06-18
Update MPI_Type_extent dummy with MPI_Type_get_extent dummy

jeremylt:
The libCEED CI pipeline is failing, and I think its because d7b18dd3a64674acac751fd68841ba1456b36e77 didn't update `mpi_dummy.f`.

See: https://travis-ci.com/github/CEED/libCEED/jobs/351063664

Not sure if `0` is a reasonable lower bound dummy value, but `lb` does not seem to be used at this time?

### DIFF
--- a/core/mpi_dummy.f
+++ b/core/mpi_dummy.f
@@ -1109,10 +1109,11 @@ c
       return
       end
 c-----------------------------------------------------------------------
-      subroutine mpi_type_extent(ikey,isize,ierr)
+      subroutine mpi_type_get_extent(ikey,lb,isize,ierr)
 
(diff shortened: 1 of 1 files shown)


jeremylt:
libCEED CI passes with this branch: https://travis-ci.com/github/CEED/libCEED/jobs/351170186

I am not sure why the Nek CI is failing though.

jeremylt:
I added d95494a to try to get the Nek5k CI passing, but it seems like it doesn't help. That commit probably can be safely removed.

jeremylt:
Thanks for fixing this bug!

PR #669, opened 2020-07-03
[tools/re2torea] correctness for reading nel/nelv

yslan:
This is a fix for conjugate heat transfer cases.


yslan:
There are more things need to correct to support CHT. 
- The write nel/nelv is wrong.
- The print of Boundary condition doesn't consider nel/nelv mismatches. 

Will make a PR after fixing them.
p.s. CI tests fails for other issues.

PR #677, opened 2020-07-25
[WIP] Separate MHD pressure solver from MHD

yslan:
This allow user to solve Poisson via semg-gmres with the crs solve stored in ifield=ifldmhd.
So, users can run Fluid solver + an additional Poisson solver.

- Support 'O' BC
- pass mask to hmh_gmres, so bpmask can be active


yslan:
The final work is to modify fcrs.c to free/re-use the slot so we can solve Poisson under different diffusion coefficients.

stgeke:
Closing this for now as there is no progress. Feel free to re-open.

PR #683, opened 2020-10-08
Set nxmp and nxmt to nxmv for steady-state case

kent0:
nxmp and nxmt are always 200, but some steady-state solutions do not converge in 200 iterations.


stgeke:
This is not required. You can adjust e.g. NMXV in usr.

PR #684, opened 2020-10-19
RANS update

dshaver14:
Updates to RANS models to include implicit terms along with a corresponding short test, also includes general access to implicit terms in the Helmholtz solver through userf and userq


stgeke:
Can you split this in 3 PRs:
- implicit passive scalar source terms
- implicit body force terms
- RANS update

dshaver14:
I can separate the RANS update no problem, but the source terms will be tougher. Both the scalar source and the body force terms are implemented using a single array that is added to h1 in sethlm. If I separate the two, it will be quite a bit more messy, but if you prefer, I can do it that way.

dshaver14:
These features will be addressed in separate pull requests.

PR #685, opened 2020-10-30
makenek resolves compiler aliases and symlinks

RonRahaman:
With these updates, makenek will resolve the aliases and symlinks of the Fortran compiler to discover the true executable.  

This update is necessitated by the mpich APT packages in the latest version of Ubuntu.  Apparantly, `mpif77` uses `f95` which, after resolving the aliases and symlinks, is in fact gfortran.  
```
$ /usr/bin/mpif77.mpich -show
f95 -Wl,-Bsymbolic-functions -Wl,-z,relro -I/usr/include/x86_64-linux-gnu/mpich -I/usr/include/x86_64-linux-gnu/mpich -L/usr/lib/x86_64-linux-gnu -lmpichfort -lmpich
$ which f95
/usr/bin/f95
$ realpath -e /usr/bin/f95
/usr/bin/x86_64-linux-gnu-gfortran-9
```

makenek doesn't recognize the vendor for `f95`, but resolving the aliases and symlinks allows makenek to proceed.

### DIFF
--- a/core/makenek.inc
+++ b/core/makenek.inc
@@ -209,6 +209,10 @@ fi
 \rm -f .tmp
 if [ $FCok -eq 0 ]; then
   FCcomp="unknown"
+else
+  # Resolve aliases and symlinks
+  FCcomp="`which $FCcomp`"
+  FCcomp="`realpath -e $FCcomp`"
 fi
 
 # assign FC compiler specific flags
--- a/tools/maketools.inc
+++ b/tools/maketools.inc
@@ -80,6 +80,10 @@ fi
 \rm -f .tmp
 if [ $FCok -eq 0 ]; then
   FCcomp="unknown"
+else
+  # Resolve aliases and symlinks
+  FCcomp="`which $FCcomp`"
+  FCcomp="`realpath -e $FCcomp`"
 fi
 
 PPPO=""


aprilnovak:
What about in other makefiles like `maketools.inc`? I had the same issue we discussed yesterday when trying to make `exo2nek`.

RonRahaman:
@aprilnovak Good point, I will fix that, too

yslan:
@RonRahaman  I found this won't work on a Cray server.

Input: `FCcomp=gfortran`
```
$ which gfortran
/opt/gcc/8.1.0/bin/gfortran
```
```
$ ls -ltra /opt/gcc/8.1.0/bin/gfortran
/opt/gcc/8.1.0/bin/gfortran -> redirect
```
```
$ realpath -e /opt/gcc/8.1.0/bin/gfortran
/cm/shared/opt/gcc/8.1.0/bin/redirect
```
It turns out that gfortran link to `/cm/shared/opt/gcc/8.1.0/bin/redirect`, which is a script.
It will redirect to another path but has to be called by gfortran
```
$ cat /cm/shared/opt/gcc/8.1.0/bin/redirect
#! /bin/sh
eval ${XTPE_SET-"set -ue"}
$(dirname $0)/../snos/bin/$(basename $0) "$@"
```

This is the full path of the real gfortran:
```
/opt/gcc/8.1.0/snos/bin/gfortran
```

RonRahaman:
Okay, another possibility is to add another case for `*f95*` (see below).  However, I don't believe that's a perfect option either, since f95 is not always gfortran.   Otherwise, this would need a more robust test, maybe by grepping `$FCcomp --version`.  Do you have any other ideas?

```
case $FCcomp in
  *pgf*)       FCPP="-Mpreprocess"
               FR8="-r8"
               ;;
  *gfortran*)  FCPP="-cpp"
               FR8="-fdefault-real-8 -fdefault-double-8"
               FF77="-std=legacy"
               ;;
  *f95*)  FCPP="-cpp"
               FR8="-fdefault-real-8 -fdefault-double-8"
               FF77="-std=legacy"
               ;;
  *ftn*)       FCPP="-eZ"
               FR8="-sreal64"
               ;;
  *armflang*)  FCPP="-cpp"
               FR8="-r8"
               FF77="-std=legacy"
               ;;
  *ifort*)     FCPP="-fpp"
               FR8="-r8 -fpconstant"
               ;;
  *xlf*)       FCPP="-qsuffix=cpp=f"
               FR8="-qrealsize=8 -qdpc=e"
               FF77="-qfixed"
               PPPO="-WF,"
               ;;
  *)  echo "ERROR: Cannot find a supported compiler!"
      echo ""
      read;;
esac

**Comment by RonRahaman** [comment]
CMake runs this test program to determine the ID, and we could probably do the same:
https://gitlab.kitware.com/cmake/cmake/-/blob/master/Modules/CMakeFortranCompilerId.F.in

**Comment by RonRahaman** [comment]
@yslan In the meantime, I submitted another PR to revert this:  https://github.com/Nek5000/Nek5000/pull/690

**Comment by yslan** [comment]
First of all, it will help recognize this issue if we print `$FCcomp`. I actually trace this from NekRS.
```
  *)  echo "ERROR: Cannot find a supported compiler!, $FCcomp"
      echo ""
```

If every compiler supports `--version`, then I vote for `$FCcomp --version`.
In such case, we don't care how complicate linking is as long as the final command is recognizable.

I also think user should be able to provide the vendor directly. 
We can always check if vendor is provided. If not, then we try to guess it from FCcomp.

We could save some popular architecture, like we have in NekCEM.
https://github.com/NekCEM/NekCEM/blob/development/bin/arch.json

RonRahaman:
I think it would be good to have an option manually specify a compiler.  And maybe NekRS can pass the vendor that it discovers.

stgeke:
You can specify the compiler using e.g. MPICH_FC.

aprilnovak:
setting `export MPICH_FC=$(which mpif90)` doesn't work for me on ulam - for some reason, `./makenrs 0 0 0` just hangs forever (this doesn't happen if I manually edit the `makenek.inc` like @RonRahaman did in this PR).

stgeke:
@aprilnovak You have to specify the Fortran compiler you want to use e.g. gfortran

aprilnovak:
oh duh, it doesn't make sense to compile MPICH with an MPI wrapper... thanks, this is a good solution for me.

PR #687, opened 2020-11-03
add new exo2nek

yhaomin2007:
1. tet-to-hex
2. periodicity
3. CHT

### DIFF
--- a/core/reader_re2.f
+++ b/core/reader_re2.f
@@ -27,7 +27,7 @@ c-----------------------------------------------------------------------
 
       ! number of fields to read
       if (param(32).gt.0) then
-        if (ifheat) then
+        if (nelgt.gt.nelgv) then
          nfldt = ibc + int(param(32))
 		else 
          nfldt = ibc + int(param(32)) - 1 
--- a/core/reader_re2.f
+++ b/core/reader_re2.f
@@ -814,7 +814,7 @@ c-----------------------------------------------------------------------
          if(version.eq.'#v002') wdsizi = 8
          if(version.eq.'#v003') then
            wdsizi = 8
-           param(32) = 1
+           !param(32) = 1
          endif
 
          call byte_read(test,1,ierr)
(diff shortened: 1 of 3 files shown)


aprilnovak:
Can this accommodate HEX8 elements now? It looks like the script still assumes HEX20 if you use option 1:

`write(6,*) '1: assume pure hex20 mesh, original exo2nek'`

It's sort of confusing that `exo2nek` (at least the previous version) doesn't allow HEX8, but from my understanding there's not an actual limitation in Nek5000 as long as the conversion tool enabled it. @stgeke can correct me if I'm wrong on this last part.

yhaomin2007:
hex8 is only accepted in option2 and option3 , which will do tet-to-hex conversion.

in the hybrid mesh, hex8 is usually generated in far field for inlet and outlet.

aprilnovak:
But since Nek always just takes the HEX8/HEX20 and constructs a higher-order HEX## based on the polynomial order, shouldn't I be allowed to create a full-HEX8 mesh? 

Correct me if I'm misunderstanding something.

yhaomin2007:
you can consider hex8 is just a linear version of hex20.

convert from hex8 to a linear hex20 is quite straight forward, and you can do this many code.

the new exo2nek will also convert hex8 to linear hex20.

aprilnovak:
Ok, so you're saying that this new exo2nek won't throw an error if my mesh is HEX8?

If so, that answers my question and works for me!

yhaomin2007:
unfortunately it will throw errors.
option2 assumes tet-wedge hybrid mesh
option3 assumes tet-hex-wedge hybrid mesh

I believe you generated your mesh in cubit.
in cubit, you can convert hex8 to hex20, but without projection to curvature.
Then your mesh mesh can be converted with option1.

aprilnovak:
Yes, I have just been making HEX20 elements in Cubit to circumvent this. It is just confusing for the user to get an error when using HEX8, because it makes me think that nekRS cannot handle HEX 8 elements (but it can).

Anyways, this is just a comment from me - as long as there is a way around it, it's not really an issue - maybe the error message could be changed from `ERROR: Only HEX20 elements are allowed in a 3D mesh!` to something like `ERROR: exo2nek can only convert HEX20 elements for 3D meshes!`.

I think that would better reflect that the limitation is in exo2nek, and not in nekRS.

PR #693, opened 2020-12-14
Removed velocity projection form vol_flow

adampep:
In v17 there were two versions of ophinv routine with and without a velocity projection. The second one was removed in v19, however this causes a problem in vol_flow calculations. It is not visible with constant dt, as a volume flow calculation is executed only once, but with a variable time step vol_flow is recalculated each time dt changes giving scary looking NaN in a log. This is caused by an attempt to enlarge projection space for inconsistent operator. Although this vector is rejected, the log is full of the error messages. This commit adds ophinv without projections for vol_flow only. Neknek version (multimesh.f) is not tested.

### DIFF
--- a/core/drive2.f
+++ b/core/drive2.f
@@ -1548,7 +1548,7 @@ C
 c
       intype = -1
       call sethlm   (h1,h2,intype)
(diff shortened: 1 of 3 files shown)


adampep:
Error message is misleading. Only tools test failed, and it is not related to changes in a pull request.

PR #706, opened 2021-01-08
Revert recent change to conv_char

yslan:
char only convects nelv


yslan:
`python -m 'unittest' NekTests.ConjHtChar.test_PnPn_Parallel`

PR #710, opened 2021-01-27
sideset untouched for one fluid exo case

yhaomin2007:
sideset untouched if there is only one fluid exo file


stgeke:
It looks like there is still an offset added to the sideset IDs in case of multiple input files. I think we should either print a warning and make them consecutive or bail out.

yhaomin2007:
Stefan, do you think there should be an option by user to say if they want sideset to be offset or not ?

yhaomin2007:
will submit a new PR later

PR #722, opened 2021-03-22
WIP: Print partition stats before and after parRSB

thilinarmtb:


stgeke:
Merged as https://github.com/Nek5000/Nek5000/commits/master

PR #725, opened 2021-03-26
Port short_tests/lib to Python >= 3.6

paugier:
It is not possible to run Nek5000 unittests without Python 2.7. CPython 2.7 is no longer maintained and Python 2.7 is not installed by default in most recent Linux desktop distributions. In 2021, it makes sense to drop Python 2.7 and to use Python >= 3.6, which is easily available in most cases.


paugier:
Hi. Is there anyone to tell me if this PR could be considered ?

PR #727, opened 2021-05-13
Running tests with GitHub Actions

RonRahaman:
This enables GitHub Actions to run the CI tests.  They are the same test cases as the Travis tests.  Currently, there are some tests failing, and I believe none of them are related to this PR:

* These are failing because the results don't match the reference solutions:
  - CmtInviscidVortex::test_Pn_Pn_Parallel
  - FsHydro::test_PnPn2_Parallel
  - KovStokes::test_PnPn2_Parallel
  - KovStokes::test_PnPn_Parallel 
* The "Tools" test is also failing with exo2nek.  There's an error in one of SEACAS' own build scripts.  I've identified the cause (SEACAS' script isn't quoting its variables properly), and it needs to be fixed in SEACAS itself.


RonRahaman:
Until this is merged in "master", the CI won't be enabled.  After it's merged, the CI will be run for PR's targeting "master" and for direct pushes to "master"

PR #730, opened 2021-06-08
Support for connectivity tol. in .par file and changes to use `next` branch of parRSB

thilinarmtb:
@stgeke: I will send a similar PR for nekRS.

### DIFF
--- a/core/map2.f
+++ b/core/map2.f
@@ -188,7 +188,7 @@ c-----------------------------------------------------------------------
         tol = connectivityTol
         call find_con(wk,nwk,tol,ierr)
         if(ierr.ne.0) then
-          tol = tol / 2.0;
+          tol = tol / 10.0;
           call find_con(wk,nwk,tol,ierr)
         endif
         call err_chk(ierr,' findConnectivity failed!$')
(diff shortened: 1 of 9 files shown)


core/map2.f
```diff
@@ -183,8 +185,12 @@ subroutine get_vert_map(nlv)
 
       if (ierr.ne.0) then
         ifread_con = .false.
-        call find_con(wk,nwk,0.2,ierr)
-        if(ierr.ne.0) call find_con(wk,nwk,0.01,ierr)
+        tol = connectivityTol
+        call find_con(wk,nwk,tol,ierr)
+        if(ierr.ne.0) then
+          tol = tol / 2.0;
```
stgeke: I think we should reduce it by 10x

thilinarmtb: Done.

thilinarmtb:
@stgeke: If there is no .par file this makes `connectivityTol` to 0. I guess I need to set defaults for when a `.rea` is present.

stgeke:
Included in https://github.com/Nek5000/Nek5000/commit/8c8f781400e05d4151f573b2e67cb30ccfb38a50

Issue #736, opened 2021-08-30
exo2nek fails for boundary condition for the TET4+WEDGE6 combination


fatihsarikurt:
TET4+WEDGE6 mesh created by ICEM CFD with defined sideset id's with inlet, outlet, and wall fails to create complete boundary conditions for re2 file. It only passes the inlet and outlet boundary sideset id's. But it does not pass the id for the wall boundary for prism layers (WEDGE6) elements.

PR #737, opened 2021-08-30
corrected variable for correct wedgess and exoss data pass issue #736

fatihsarikurt:
this change solves the problem, tested for the TEX+WEDGE combination, it works


stgeke:
Fixed in https://github.com/Nek5000/Nek5000/commit/d8d6f9cb497cff724c204d72015e7a22ddf57d34

Issue #729, opened 2021-05-27
Name conflicts with Cray LibSci 


stgeke:
If LibSci gets loaded (at runtime) it overwrites conflicting symbols like icopy. This results in a SEGV. We should rename the conflicting symbols. 

Current workaround:  Ensure that LibSci is not loaded.

MalachiTimothyPhillips:
Subroutines to change (looking through `math.f`):

- [ ] icopy
- [ ] cmult
- [ ] iswap
- [ ] isort
- [ ] sort

There may be more.

PR #747, opened 2021-11-09
[WIP] Rename subroutine to avoid cray libsci name conflicts (#729)

MalachiTimothyPhillips:
- `icopy` -> `intcopy`
- `cmult` -> `constMult`
- `iswap` -> `intswap`
- `isort` -> `intsort`
- `sort`  -> `realSort`


stgeke:
Renaming these functions will break user code. For now I cannot think of a good solution to fix this. We're rely on the user to unload the libsci module before nek is getting compiled.

PR #748, opened 2021-11-26
Write PARALLEL and mpif.h in CASEDIR

ashwinvis:
Writing into the core directory triggers a race conditions when multiple
compilations are launched simultaneously. These essential include files
would vanish during compilation. As a fix, instead of copying "in-place"
within the ``core`` directory, the files are copied to CASEDIR.

This is the simplest approach I could think of. See also: https://github.com/exabl/snek5000/pull/117 where we discovered this. Note that this is not related to snek5000 in any sense, and can happen if we simply execute `makenek` simultaneously multiple times from different directories.

### DIFF
--- a/core/makefile.template
+++ b/core/makefile.template
@@ -75,10 +75,10 @@ else
    filters_cmt.o diagnostics.o artvisc.o
 endif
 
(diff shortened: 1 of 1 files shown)


ashwinvis:
Rebased in #749

PR #749, opened 2021-11-26
Write PARALLEL and mpif.h in OBJDIR

ashwinvis:
Writing into the core directory triggers a race conditions when multiple compilations are launched simultaneously. These essential include files would vanish during compilation. As a fix, instead of copying "in-place" within the core directory, the files are copied to CASEDIR.

This is the simplest approach I could think of. See also: exabl/snek5000#117 where we discovered this. Note that this is not related to snek5000 in any sense, and can happen if we simply execute makenek simultaneously multiple times from different directories.

### DIFF
--- a/core/makefile.template
+++ b/core/makefile.template
@@ -64,10 +64,10 @@ ifeq ($(MPI),0)
   COMM_MPI := ${COMM_MPI} mpi_dummy.o
 endif
 
-DUMMY:= $(shell cp $S/core/PARALLEL.default $S/core/PARALLEL 2>/dev/null)
+DUMMY:= $(shell cp $S/core/PARALLEL.default $(CASEDIR)/PARALLEL 2>/dev/null)
 ifeq ($(DPROCMAP),1)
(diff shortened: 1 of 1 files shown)


paugier:
Are there any Nek5000 developers to review these changes and give feedback?

stgeke:
Please write into CASEDIR/obj instead of just CASEDIR.

ashwinvis:
@stgeke Done in ffe18cc

stgeke:
Please add OBJDIR as an INCLUDE path otherwise it won't compile.

stgeke:
Please ensure that your PR passes CI testing.

stgeke:
Added in https://github.com/Nek5000/Nek5000/commit/3e538555256b047a6d534111efa669e0557d0979

ashwinvis:
Looks good, but I should point out that https://github.com/Nek5000/Nek5000/commit/3e538555256b047a6d534111efa669e0557d0979 missed changing the location for `mpif.h`.

PR #751, opened 2022-01-27
Add interpolation as a restart option

dshaver14:
Adds a simple parameter for setics to call gfldr instead of restart.


stgeke:
I don't think a new key in par is required. Interpolation should just be a restart option similar to e.g. `t=0` or `UPT`.

dshaver14:
so, something like
`startFrom = case0.f00001 int`?

stgeke:
Yeah

PR #752, opened 2022-01-30
Add 'int' restart option for interpolation

dshaver14:
calls gfldr for restart when using
`startFrom = case0.f00001 I` 

addresses #415 

currently only supported if param(67) = 6

### DIFF
--- a/core/RESTART
+++ b/core/RESTART
@@ -31,10 +31,10 @@ c
 
       logical ifgetx ,ifgetu ,ifgetp ,ifgett ,ifgtps (ldimt1),ifgtim
      $       ,ifgetxr,ifgetur,ifgetpr,ifgettr,ifgtpsr(ldimt1),ifgtimr
-     $       ,if_byte_sw,ifgetz,ifgetw,ifdiro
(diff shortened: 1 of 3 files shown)


stgeke:
Can we have something more descriptive then just 'i' e.g. 'int'.

stgeke:
The CI tests for this PR fail. Please have a look.

dshaver14:
I can add a dummy gfldr subroutine that's defined for NOMPIIO that will fix the EddyUv failed test, but when I try to run  InclDef::test_PnPn_Parallel, I get 

`ABORT: Too many processors (       2) for to few elements (           1).`

This fails with the master branch too. Is this normal? how does this test ever pass at all?

yslan:
The InclDef only has 1 element, so it will fail if you run with 2 processors.

Interestingly, it is set to be compiled with IFMPI=0 and `-DNOMPIIO` 
For other passed test, there are two independent serial jobs launched at the same time.
https://github.com/Nek5000/Nek5000/runs/4998981918?check_suite_focus=true

dshaver14:
So my real issue here is that I assumed the CI tests would be identical to the short_tests. I'll open a separate issue for that. In any case, I expect my changes should pass now.

PR #765, opened 2022-08-30
[WIP] Make max path length 1024 characters, instead of 132 characters.

### LINKED ISSUE #763: Allow for path names longer than 132 characters

MalachiTimothyPhillips:
Closes #763


stgeke:
I understand the motivation of this PR but I am afraid this change will break many existing usr file. That's not a good idea. Closing this PR now until we find a intrusive fix.

PR #767, opened 2022-09-05
1 billion elements support

misunmin:
* E={1024900}x{9,10,100,800,1000} tested
 * n2to3: (E>=10M, #v004, i16), (E<10M, #v002, i9), backward compatible.
 * n2to3co2: (E>=10M, #v002, i16), (E<10M, #v001, i12), backward compatible.
 * gencon, genbox, postnek, prenek: no E=1B support

 Co-authored-by: Paul Fischer, Yu-Hsiang Lan

### DIFF
--- a/core/map2.f
+++ b/core/map2.f
@@ -397,8 +397,14 @@ c-----------------------------------------------------------------------
             call byte_read(hdr,sizeof(hdr)/4,ierr)
             if(ierr.ne.0) goto 100
 
-            read (hdr,*) version,nelgti,nelgvi,nvi
-c    1       format(a5,2i12,i2)
+            read (hdr,'(a5)') version 
+
+            if (version.eq.'#v002') then
+               read (hdr,*) version,nelgti,nelgvi,nvi
+            else
+               read (hdr,1) version,nelgti,nelgvi,nvi
+            endif
+            write (6,*) ' hdr calling ',version,nelgti,nelgvi,nvi 
 
             call byte_read(test,1,ierr)
             if(ierr.ne.0) goto 100
@@ -406,6 +412,9 @@ c    1       format(a5,2i12,i2)
             if(ierr.ne.0) goto 100
          endif
       endif
+
+   1  format(a5,3i12)
+
       call bcast(nelgti,sizeof(nelgti))
       call bcast(nelgvi,sizeof(nelgvi))
       call bcast(nvi,sizeof(nvi))
--- a/core/reader_re2.f
+++ b/core/reader_re2.f
@@ -90,10 +90,10 @@ c-----------------------------------------------------------------------
       common /ctmp1/  vi
 
       integer*8       lre2off_b,dtmp8
-      integer*8       nrg
+      integer*8       nrg      
 
       nrg       = nelgt
-      nr        = nelt
+      nr        = nelt  ! 09/03/2022, supports only up to 2B elements
       irankoff  = igl_running_sum(nr) - nr
       dtmp8     = irankoff
       re2off_b  = 84 ! set initial offset (hdr + endian)
@@ -165,7 +165,7 @@ c-----------------------------------------------------------------------
       common /ctmp1/  vi
 
       integer*8       lre2off_b,dtmp8
-      integer*8       nrg
+      integer*8       nrg,nr
       integer*4       nrg4(2)
      
       integer*8       i8gl_running_sum 
@@ -202,7 +202,7 @@ c-----------------------------------------------------------------------
       re2off_b = re2off_b + nrg*4*lrs4
 
       if (.not.ifread) return
-      if(nio.eq.0) write(6,'(A,I10)') ' reading curved sides   ', nrg
+      if(nio.eq.0) write(6,'(A,I20)') ' reading curved sides   ', nrg
 
       nwds4r = nr*lrs4
       call byte_set_view(lre2off_b,fh_re2)
@@ -269,7 +269,7 @@ c-----------------------------------------------------------------------
       common /ctmp1/  vi
 
       integer*8       lre2off_b,dtmp8
-      integer*8       nrg
+      integer*8       nrg,nr
       integer*4       nrg4(2)
 
       integer*8       i8gl_running_sum 
@@ -306,7 +306,7 @@ c-----------------------------------------------------------------------
       re2off_b = re2off_b + nrg*4*lrs4
 
       if (.not.ifread) return
-      if(nio.eq.0) write(6,'(A,I10,A,I3)') 
+      if(nio.eq.0) write(6,'(A,I20,A,I3)') 
      $             ' reading boundary faces ', nrg, 
      $             ' for ifield ', ifield
 
@@ -550,6 +550,7 @@ c-----------------------------------------------------------------------
       include 'SIZE'
       include 'TOTAL'
       logical ifbswap
+      integer*8 ncurve8
 
       integer e,eg,buf(55)
       real rcurve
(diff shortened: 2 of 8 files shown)


stgeke:
May I suggest a few TODOs:
- [ ] Add number of BC fields to re2 hdr `#004` and adjust param(32)
- [ ] Eliminate co/co2 file support (no longer required as we have parCon)
- [ ] Add large elements support for genbox

core/map2.f
```diff
@@ -397,15 +397,24 @@ subroutine read_con(wk,nwk,nelr,nv,ierr)
             call byte_read(hdr,sizeof(hdr)/4,ierr)
             if(ierr.ne.0) goto 100
 
-            read (hdr,*) version,nelgti,nelgvi,nvi
-c    1       format(a5,2i12,i2)
+            read (hdr,'(a5)') version 
+
+            if (version.eq.'#v002') then
```
stgeke, line 402: Why do we need a new header version `#v002` if the old one was using free format already?

core/reader_re2.f
```diff
@@ -90,10 +90,10 @@ subroutine readp_re2_mesh(ifbswap,ifread) ! version 2 of .re2 reader
       common /ctmp1/  vi
 
       integer*8       lre2off_b,dtmp8
-      integer*8       nrg
+      integer*8       nrg      
 
       nrg       = nelgt
-      nr        = nelt
+      nr        = nelt  ! 09/03/2022, supports only up to 2B elements
```
stgeke, line 96: I find this comment confusing. The 2B element restriction isn't introduced here.

core/map2.f
```diff
@@ -397,15 +397,24 @@ subroutine read_con(wk,nwk,nelr,nv,ierr)
...
+               read (hdr,*) version,nelgti,nelgvi,nvi
+            else
+               read (hdr,1) version,nelgti,nelgvi,nvi
+            endif
+            write (6,*) ' hdr calling ',version,nelgti,nelgvi,nvi
```
stgeke, line 407: Is this a left over debug statement?

core/reader_re2.f
```diff
@@ -814,15 +816,25 @@ subroutine read_re2_hdr(ifbswap, ifverbose) ! open file & chk for byteswap
...
          wdsizi = 4
          if(version.eq.'#v002') wdsizi = 8
-         if(version.eq.'#v003') then
-           wdsizi = 8
-           param(32)=1
-         endif
+         if(version.eq.'#v003') wdsizi = 8
+         if(version.eq.'#v004') wdsizi = 8
+
+         if(version.eq.'#v003') param(32)=1 !param(32)=0 read all BC from .re2
```
stgeke, line 837: Ethier we introduce with `#v004` the number of BC fields in the header and adjust param(32) or we assume that from `#v003` on there is only one boundary field.

tools/n2to3co2/n2to3co2.f
```diff
@@ -0,0 +1,742 @@
+c----------------------------------------------------------------------
+      program n2to3co2
```
stgeke, line 2: Why do we need n2to3co2 and the regular n2to3?

stgeke (changes requested):

core/reader_re2.f
misunmin: nr is not integer*8 in this routine and so it was a reminder note for future changes, but not necessary to keep it here. We can remove.

core/map2.f
misunmin: Necessary to print the header info right away -  critical for large problems while nelgt/nelgv/lelt is printed long after parCon & parRSB. Cosmetic can change if 'hdr calling' looks like debug stat. Suggestion?

tools/n2to3co2/n2to3co2.f
misunmin: n2to3co2 is co2 generator using i16. Keeping this tool separate from n2to3 is simpler instead of putting two functions into one tool, n2to3. At least, trying to keep 1B support version to be as minimal change as possible without much deviating from previous version in terms of format and style.

misunmin:
We can allow two calls of "read_re2_data"  at least in this version - to minimize the changes. More changes can go to 2B.

core/reader_re2.f
misunmin: I simply kept the previous setting:  #v003 + param(32)=1, to avoid potential incompatibility.  
We can add nBCre2 if preferred. But this will involve more touches in re2 generating routines, needing another 20000-30000 node hours to test.

misunmin:
Removal of one of the two calls of "read_re2_data" will require another 20000-30000 node hours for testing large cases. I suggest next round (2B) to include further significant changes.

core/reader_re2.f
stgeke: `nr` stores the local number of "records" (here elements) to read. No need for `integer*8`

core/reader_re2.f
stgeke: `#v004` is changing the previous setting because it doesn't enforce `param(32) = 1` like `#v003`. This is confusing.

tools/n2to3co2/n2to3co2.f
stgeke: n2to3co2 is more or less a copy of n2to3 with a minor extension. I prefer to avoid any kind of code duplication unless there is a very good reason to do so. For now I am ok to have an extra version. Medium term co2 should be replaced by parCon anyway.

stgeke:
Where does the 1 billion element restriction come from? Currently there seem to be two limits:
1. vertex IDs in ma2/co2 are stored as `integer*4`
2. global elements IDs are stored as `integer*4`

The first one is more restrictive and will be mesh specific. In an ideal (periodic) cube-box case (number of unique vertices ~ number of elements) it translates to ~2B elements.

core/reader_re2.f
misunmin: Yes.

core/reader_re2.f
misunmin: #v004 header is created by only n2to3.

core/reader_re2.f
misunmin: nr        = nelt  ! 09/03/2022, supports only up to 2B elements
      irankoff  = igl_running_sum(nr) - nr

The note was for the next line -- eventually needing irankoff8 = i8gl_running_sum(nr) - nr
with nr=nelt where nr integer*8 for big size problems.

core/map2.f
misunmin: All co2 #v001 using i12 generated from existing tools will not get affected anyway.

In n2to3co3: we still want to support co2 generation with #v001, but for E<10M. We set E=10M as the transition criteria to big size (10M is to have more users for tests at this size. 100M-1B have lesser chance). For E>=10M, the header uses #v002 with i16.  The generation of re2 (#v004) is based on the same approach, supporting #v002 and #v004 using E=10M as the transition to big.

And so the decision #v002 for co2 was for the consistency with re2 using i16 and E=10M criteria, indicating the jump to 1B support. 

Theoretically, things will have to work up to 2B. More than a dozen tests were done with 1B level. The range of 1B-2B has to be done more testings to claim 2B support with additional changes at next round.

core/reader_re2.f
stgeke: Got it. Indeed this has to change but there are many other locations which need to be changed too. So having a comment just for this line can be misleading. Anyway it's a minor issue.

misunmin:
> Where does the 1 billion element restriction come from? Currently there seem to be two limits:
> 
> 1. vertex IDs in ma2/co2 are stored as `integer*4`
> 2. global elements IDs are stored as `integer*4`
> 
> The first one is more restrictive and will be mesh specific. In an ideal (periodic) cube-box case (number of unique vertices ~ number of elements) it translates to ~2B elements.

1B support doesn't mean it is restricted to 1B. It's more about verification done for 1B support. More than a dozen tests were done with 1B level. The range of 1B-2B has to be done more testings to claim 2B support with additional changes at next round. We can update the commit message if it sounds like 1B restriction.

core/reader_re2.f
misunmin: Yes too many places to change accordingly once int*8 applied, and so we kept the routine as they were, but leaving a small hint to remember. We can change the comment as preferred or I'm fine to remove.

stgeke: I suggest to remove the comment but create an detailed GitHub issue what remaining issue we need to fix.

stgeke:
Replaced by #768

PR #776, opened 2022-10-20
Changes for new parRSB

thilinarmtb:


core/map2.f
```diff
@@ -186,12 +186,13 @@ subroutine get_vert_map(nlv)
       if (ierr.ne.0) then
         ifread_con = .false.
         tol = connectivityTol
-        call find_con(wk,size(wk),tol,ierr)
+        call find_con(wk,nwk,tol,1,ierr)
         if(ierr.ne.0) then
           tol = tol / 10.0;
-          call find_con(wk,size(wk),tol,ierr)
+          call find_con(wk,nwk,tol,1,ierr)
```
thilinarmtb, line 192: @stgeke : Is this change (`nwk` instead of `size(wk)`) correct?

thilinarmtb: Line 189 has a similar change as well.

stgeke:
Merged in https://github.com/Nek5000/Nek5000/commit/165d946c97f4848c6d91a7e6ec561de6b2b9e16a

PR #778, opened 2022-11-06
Adding user parameters in PARDICT file

akhoubani:
Hello,

In developing our solver with [snek5000](https://github.com/snek5000), we noticed that while it is thought of 20 user parameters inside the NEK, only 10 user parameters are defined in [PARDICT](https://github.com/snek5000/Nek5000/blob/140c4fd96c28df5f4d415e89be57424a7eba3691/core/PARDICT) file. I wonder if you could consider this pull request.

Kind regards.


stgeke:
Can you please group all user params together.

PR #781, opened 2022-11-10
Add flow rate settings for par file

joneuhauser:
The flow rate settings in param(54) and param(55) were previously not available via the par file. 

The names of the parameters are derived from the corresponding settings for NekRS.

This PR adds those settings, as well as some validation on the parameters. Apart from the indices in PARDICT, the patch can be applied to v19.0 if someone needs it (already tested).

New syntax:
```
[GENERAL]
constFlowRate = <[XYZ]>
meanVolumetricFlow = <float>
OR
meanVelocity = <float>
```

Until then, the parameters can be set with a workaround: 

```
case.par:

[General]
userParam01 = -3        # fixed flow rate dir
userParam02 = 1         # vol flow rate (up05>0) or Ubar (up05<0)

case.usr, subroutine usrdat3:

param(54) = uparam(1)
param(55) = uparam(2)

```

Should I open a MR for the documentation repository as well, or only after this is released?

### DIFF
--- a/core/PARDICT
+++ b/core/PARDICT
@@ -4,7 +4,7 @@ c     Note: Keys have to be in captial letters
 c
       integer PARDICT_NKEYS
 
-      parameter(PARDICT_NKEYS = 104)
+      parameter(PARDICT_NKEYS = 107)
 
       character*132 pardictkey(PARDICT_NKEYS)
       data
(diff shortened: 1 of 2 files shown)


dshaver14 (changes requested):
Please pull the latest master to update the number of PARDICT keys. Then this will be ready to merge.

> and i suggest changing the syntax to `constantFlowRate` instead of `constFlowRate`

Ignore this suggestion, the original is consistent with NekRS.

dshaver14:
@joneuhauser if you also want to update the NekDoc repo to reflect this change, I would appreciate it.

dshaver14:
covered by #791

PR #789, opened 2023-01-08
update readme files, and add CHT to gmsh2nek

yhaomin2007:


stgeke:
Why does this PR touch many more file than just the readme?

yhaomin2007:
All commits except the last one are made for the previous pull request, which is already merged. Only the last commit is for this pull request.

stgeke:
The PR should only change the readme.

yhaomin2007:
Yes, i know. but the CHT feature in gmsh2nek is also request by some users.

PR #794, opened 2023-02-05
correct FFLAGS used in tools

yslan:
- Use FFLAGS instead of BIGMEM    
  so the FFLAGS in maketools will be applied when compiling the tools.

- Ignore the argument-mismatch error in cgns2nek    
  The newer Fortran compiler turns this warning to error.


yslan:
this should fix the CI / Tools error

PR #799, opened 2023-03-23
Fix incorrect handling of env-var NEK_LOGFILE

### LINKED ISSUE #798: Correct boolean statement
I am getting a compiler warning about the following line in `core/chelpers.c`: 

```
  else if (envvar = getenv("NEK_LOGFILE")) {
    if (*sid >= 0) sprintf(logfile, "s%05d_", *sid);
    strcat(logfile + strlen(logfile), envvar);
    redirect = 1;
  }
```

That should be a `==` instead of a `=` inside the if statement. I don't think this is causing any bad behavior for my case, but wanted to note it because it's probably a mistake.

aprilnovak:
Closes #798

### DIFF
--- a/core/chelpers.c
+++ b/core/chelpers.c
@@ -74,12 +74,12 @@ void set_stdout(char *f, int *sid, int flen)
   logfile[i+1] = '\0';
 
   int redirect = 0;
-  char *envvar;
 
   if (logfile[0] != '\0') {
     redirect = 1;
   } 
-  else if (envvar = getenv("NEK_LOGFILE")) {
+  else if (getenv("NEK_LOGFILE")) {
+    char *envvar = getenv("NEK_LOGFILE");
     if (*sid >= 0) sprintf(logfile, "s%05d_", *sid);
     strcat(logfile + strlen(logfile), envvar);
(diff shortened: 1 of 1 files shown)


core/chelpers.c
```diff
@@ -79,7 +79,7 @@ void set_stdout(char *f, int *sid, int flen)
   if (logfile[0] != '\0') {
     redirect = 1;
   } 
-  else if (envvar = getenv("NEK_LOGFILE")) {
+  else if (envvar == getenv("NEK_LOGFILE")) {
```
stgeke: I think this block should read:
```
else if (getenv("NEK_LOGFILE")) {
    char *envvar = getenv("NEK_LOGFILE");
    if (*sid >= 0) sprintf(logfile, "s%05d_", *sid);
    strcat(logfile + strlen(logfile), envvar);
    redirect = 1;
}

```

stgeke (changes requested):

core/chelpers.c
aprilnovak: You're probably right, I didn't look too closely at what this code does.

PR #805, opened 2023-04-24
[maketool] update the script with a better checking

yslan:
- check tool from the available list
  If fails, it prints something like this

  ```
  $ ./maketools tt
  FATEL ERROR: "tt" is not a tool
  (avaliable list) genmap gencon genbox n2to3 reatore2 nekmerge prenek postnek nekamg_setup gmsh2nek exo2nek cgns2nek n2to3co2
  ```
- check dependencies of each tool independently
  ```
  function chk_web  # openssl, wget, tar
      cgns2nek
      exo2nek
      nekamg_setup
  function chk_cmake 
      cgns2nek
      exo2nek
      nekamg_setup
  function chk_x11() {
      prenek
  ```

This is not well-tested yet
TODO: test the checks in the env that doesn't have cmake or x11, etc.

### DIFF
--- a/tools/maketools.inc
+++ b/tools/maketools.inc
@@ -147,29 +147,52 @@ export R8
 export CPPF
 export BIGMEM
 export LDFLAGS
-
-if ! which openssl >/dev/null; then
-  echo "FATAL ERROR: Cannot find openssl!"
-  exit 11
-fi
-if ! which wget >/dev/null; then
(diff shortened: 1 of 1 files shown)


stgeke:
@yslan Is this ready to be merged or is more testing needed?

yslan:
I don't know how to find the environment without wget or tar but the cmake check is working ok.

PR #807, opened 2023-05-19
Fix mesh_metrics's spacing

yslan:
My understanding is, we approximate each element with N^d uniform boxes with total volume = volume of the element.
Therefore, the 1D length scale should be square root for 2D.

### DIFF
--- a/core/genxyz.f
+++ b/core/genxyz.f
@@ -1496,7 +1496,7 @@ c-----------------------------------------------------------------------
       do ie = 1,nelt
          ddmin = min(ddmin,dxmin_e(ie)) 
          ddmax = max(ddmax,dxmax_e(ie))
-         dtmp  = vlsum(JACM1(1,1,1,ie),nxyz)**1./ldim
+         dtmp  = vlsum(bm1(1,1,1,ie),nxyz)**1./ldim
          ddavg = ddavg + dtmp/(lx1-1) 
       enddo 
       dxmin = glmin(ddmin,1)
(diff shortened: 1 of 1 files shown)


core/genxyz.f
```diff
@@ -1496,7 +1496,7 @@ subroutine mesh_metrics
       do ie = 1,nelt
          ddmin = min(ddmin,dxmin_e(ie)) 
          ddmax = max(ddmax,dxmax_e(ie))
-         dtmp  = vlsum(JACM1(1,1,1,ie),nxyz)**1./3
+         dtmp  = vlsum(bm1(1,1,1,ie),nxyz)**1./ldim
```
stgeke: Why was `bm1` changed to `JACM1`?

yslan: Jacobian determinant is a scale from changing the variables. It doesn't have unit.
The sum of bm1 is volume, and the cubic root of it is the length scale.

Take one element = [-1,1]^3 as an example,
jacm1 are all ones so
```
vlsum(JACM1(1,1,1,ie),nxyz)**1./3 / (lx1-1) = lx1 / N
```
On the other hand, 
```
vlsum(bm1(1,1,1,ie),nxyz)**1./3  / (lx1-1)  = 2 / N
```
which is what we expect for [-1,1] in N intervals.

We didn't print this number out anyway. I just found this inconsistency while checking the code...

stgeke: If it's not used we should simply remove it.

PR #813, opened 2023-07-06
updated gmsh2nek and exo2nek to create v004 mesh

yhaomin2007:


dshaver14:
We should revise this. I would recommend always doing the non-right-hand check and just print the number of "suspect" elements, then prompt the user if they want to fix any non-right-hand elements.

yhaomin2007:
forced non-right-hand check now, and will print the number of non-right-hand elements.

yhaomin2007:
any further comments?

stgeke:
@dshaver-ANL Is this ready to be merged?

stgeke:
I think this PR should also update cgns2nek.

yhaomin2007:
cgns2nek is also changed for v004 mesh.

dshaver14:
one more minor change, if it doesn't detect any non-right-handed elements, it shouldn't ask the user if it wants to fix them.

yhaomin2007:
changes are done. thanks for the comment.

tools/gmsh2nek/gmsh2nek.f90
```diff
@@ -1934,17 +1941,20 @@ subroutine open_re2
...
+
+      nBCre2 = 1
+      if (num_elem.ne.eftot)  nBCre2  = 2
+
 !  Write the header
       call blank     (hdr,80)    
-      write(hdr,1) num_elem, num_dim, eftot
-    1 format('#v002',i9,i3,i9,' this is the hdr')
+      write(hdr,1) num_elem, num_dim, eftot, nBCre2
+    1 format('#v004',i9,i3,i9,i3,' this is the hdr')
```
stgeke: I don't think the format matches the one we're using for reading.

yhaomin2007: fixed

stgeke:
The format is still not correct.
See here: https://github.com/Nek5000/Nek5000/blob/master/core/reader_re2.f#L585

Please fix all mesh converters including cgns2nek.

yhaomin2007:
write(hdr,1) num_elem, num_dim, eftot, nBCre2
    1 format('#v004',i16,i3,i16,i4,' hdr')
this is consistent with reader_re2, and also consistent with 
https://github.com/Nek5000/Nek5000/blob/11e94661faa8991b5c6e6711daf89d1cf1aa3ed5/tools/n2to3/n2to3.f#L334

I also chagned this in cgns2nek

dshaver14 (approved):
This looks good now. 
@stgeke Note that I had to remove the `-fallow-argument-mismatch` to get cgns2nek to compile. I think it's an incompatibility with the version of gcc I'm using (9.3.1), but a quick google search indicates that it's redundant with `-std=legacy`.

yhaomin2007:
let me know if there is anything else you want me to change. thanks

Issue #806, opened 2023-05-19
ioinfo not working


cfrouzak:
In the latest comm_mpi.f, the path is blanked out:

 128       !path = path_mult(idsess)
 129       call blank(path,len(path))

and the content of the ioinfo file in subroutine check_ioinfo (prepost.f)  is not checked.

Should it be changed to:

 128       path = path_mult(idsess)
 129       !call blank(path,len(path))

stgeke:
Fixed in https://github.com/Nek5000/Nek5000/commit/354d9824f5a39bd7e1d3906a31eb4f8260a5a9e3

PR #816, opened 2023-08-03
Fix ioinfo

dshaver14:
Addresses issue #806.

I think the real problem is that drive2 sets PATH to '/', when it should remain blank, but I'm not sure. This at least seems to fix ioinfo for now.


stgeke:
Fixed in https://github.com/Nek5000/Nek5000/commit/354d9824f5a39bd7e1d3906a31eb4f8260a5a9e3

PR #818, opened 2023-09-26
update gencon/genmap to read v004 mesh.

yhaomin2007:
update gencon/genmap to read v004 mesh.


stgeke:
How about `#v001`?

yhaomin2007:
v001 is also considered now.

Issue #834, opened 2024-07-17
A better support of v004 format, reader and writer for tools


yslan:
- v004
  Both re2torea and reatore2 don't support the v004 format.
  However, exo2nek, gmsh2nek and cgns2nek are now writing v004.
  We should extend the support of v004.

- lib for reader and writer
  All of our tools requires either a mesh reader or writer, we should gather them to the same location.
  This can easily extend the support of re2 for many tools, such as n2to3, prenek, etc.

PR #835, opened 2024-07-17
Add v004 support for reatore2

### LINKED ISSUE #795: re2torea: need an check to detect input filename = output filename
If we accidentally use the same name for the input and output, it will overwrite the rea file and corrupt the file. The tool also run to the end without error message.

A simple check would be nice.

To reproduce:
``` 
$ ls 
t.re2  t.rea

$ re2torea
Input old .rea and .re2 (source) file name:
t
 Input new .rea (output) file name:
t
 byte swap: F   6.54321003      -2.93127722E+35
         192           3         192
 number of curved sides           0
           1         224 number of bc
```

The reatore2 has the same issue but it will at least give a runtime error like this
```
Fortran runtime error: End of file
```

yslan:
This is a partial solution to https://github.com/Nek5000/Nek5000/issues/834 and also close https://github.com/Nek5000/Nek5000/issues/795

### DIFF
--- a/tools/reatore2/re2torea.f
+++ b/tools/reatore2/re2torea.f
@@ -38,7 +38,7 @@ c     an ascii rea file for just the parameters
 
       real*8 rbc,rcurve
       character*3 cbc(6,nelm)
-      logical ifbswap,if_byte_swap_test,ifexist
+      logical ifbswap,if_byte_swap_test,ifexist,strcmp
       
 c     for workstation:
       in = 5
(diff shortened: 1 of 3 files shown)


yslan:
Instruction to convert `re2` file for v004 format

1. `head -c 132 a.re2|strings` to confirm it's v004, get dimension, and nBC
    For example, the below line has dim=3, nBC=2
    ```
   #v004           10496  3            8448   2 hdr
    ```
2. Copy a dummy rea file, for example, `Nek5000/short_tests/chan2d/chan2d.rea`. 
    Rename it to `a.rea`. Change the dimension to 3 to match the re2, `  3  DIMENSIONAL RUN`
    One also needs to ensure the nBC is correct in `a.rea`. Modify `param(32)` (nBC) correspondingly.

3. call `re2torea`, feed `a` as input filename then `b` as output filename

PR #856, opened 2025-06-08
Distribute read elem via crystal router + batches

yslan:
Add an option to distribute data using crystal router.

```
! INPUT:
ifcrrs = .false. ! T = switch from rdma to crystal router
lbrst = lelt ! block size for batched restart
```

### DIFF
--- a/core/drive2.f
+++ b/core/drive2.f
@@ -54,6 +54,7 @@ c     Set default logicals
       if (lx1.eq.lx2) ifsplit=.true.
 
       if_full_pres = .false.
+      ifcrrs = .false. ! crystal router restart
 
       CALL RZERO (PARAM,200)
 
--- a/core/drive2.f
+++ b/core/drive2.f
@@ -55,7 +55,7 @@ c     Set default logicals
 
       if_full_pres = .false.
       ifcrrs = .false. ! crystal router restart
-      lbrst = lelt ! batch size for restart
+      lbrst = min(1024, lelt) ! batch size for restart
 
       CALL RZERO (PARAM,200)
 
--- a/core/drive2.f
+++ b/core/drive2.f
@@ -55,6 +55,7 @@ c     Set default logicals
 
       if_full_pres = .false.
       ifcrrs = .false. ! crystal router restart
+      lbrst = lelt ! batch size for restart
 
       CALL RZERO (PARAM,200)
 
--- a/core/INPUT
+++ b/core/INPUT
@@ -50,7 +50,7 @@ c
      $               ,ifcyclic,ifmoab,ifcoup, ifvcoup, ifusermv,ifreguo
      $               ,ifxyo_,ifaziv,ifneknek,ifneknekm,ifneknekc
      $               ,ifcvfld(ldimt1),ifdp0dt
-     $               ,ifmpiio,ifrich,ifvvisp
+     $               ,ifmpiio,ifiocrystal,ifrich,ifvvisp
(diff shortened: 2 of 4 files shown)


yslan:
Maybe put `ifiocrystal` and `fgslib_crystal_setup` to an earlier location?

yslan:
TODO: Add batched version in either ic.f or gslib to avoid n > INT_MAX.

stgeke:
Please add batched version for both RDMA and CR.

yslan:
The batches are determined based on the jeln = local element id on the target MPI rank.

```
er(e) ! global element id that just read by this rank
jnid = gllnid(er(e)) ! mpi rank of the destination
jeln = gllnel(er(e)) ! local element id of the destination rank


nbatch = (nelrr - 1) / lbrst + 1 
  do ibatch = 1,nbatch
    jeln1 = (ibatch-1)*lbrst+1
    jeln2 = ibatch*lbrst

...

    jeln = gllel(er(e))
    if (jeln.ge.jeln1.AND.jeln.le.jeln2) then
     ! participate rdma or crystal router.
    endif

```

yslan:
The batched version hasn't been tested.

yslan:
The batch version has bug. Looking into it right now.

yslan:
Accuracy test is passed

np=2

```
ERROR: fp32 rdma         0.2169516E-07  0.2334414E-07  0.2592083E-07  0.2438569E-07  0.2438710E-07
ERROR: fp32 rdma batch   0.2169516E-07  0.2334414E-07  0.2592083E-07  0.2438569E-07  0.2438710E-07
ERROR: fp32 cr           0.2169516E-07  0.2334414E-07  0.2592083E-07  0.2438569E-07  0.2438710E-07
ERROR: fp32 cr batch     0.2169516E-07  0.2334414E-07  0.2592083E-07  0.2438569E-07  0.2438710E-07
ERROR: fp64 rdma         0.6818794E-18  0.2534935E-18  0.8447292E-18  0.4717812E-14  0.1358258E-17
ERROR: fp64 rdma batch   0.6818794E-18  0.2534935E-18  0.8447292E-18  0.4717812E-14  0.1358258E-17
ERROR: fp64 cr           0.6818794E-18  0.2534935E-18  0.8447292E-18  0.4717812E-14  0.1358258E-17
ERROR: fp64 cr batch     0.6818794E-18  0.2534935E-18  0.8447292E-18  0.4717812E-14  0.1358258E-17
```

yslan:
@stgeke This passes all of my tests.

E=1M, N=7, on Frontier (CPU) 2, 3, 4, 20 nodes with 56 ranks per node.
- block size ranges from 32 to min(16384,lelt)
- Both fp32 and fp64
- Both cr and rdma

yslan:
I propose to set lbrst = 1024 which will work for most cases without losing much performance.

rdma rely on vendor MPI but generally faster.
cr is more robust but can have huge traffic, which is not shown in this plot. However, for nekrs, we live at nelt > 8000 so generally setting lbrst = 1024 should at least survive 8x larger range.


compare

core/drive2.f
```diff
@@ -54,6 +54,8 @@ subroutine initdat
       if (lx1.eq.lx2) ifsplit=.true.
 
       if_full_pres = .false.
+      ifcrrs = .false. ! crystal router restart
+      lbrst = lelt ! batch size for restart
```
stgeke: I suggest to use

```
lbrst = min(1024, lelt)
```

PR #862, opened 2025-09-12
h-refine

yslan:
Global refinement divides the edges of each element into `Ncut` equally spaced segments, producing a finer mesh with $E_{new} = N{cut}^{ldim} \times E_{old}$

**Features**
1. Recursive action before `usrdat2` (See `drive1.f`)    
    Usage: In par, `[MESH] hrefine = 2,3`
    Users can also call `hrefine_usrdat2(ncut)` manually.

2. Support restart (See `ic.f`) by recording schedule into checkpoint file
    **New checkpoint header format** is introduced:
    - Write a file
       - The h-schedule will be saved into header using 4 characters (5 spaces) after `if_full_pres`.
       - Each character represents a round of refine, support up to 4 rounds, stop at first zero.
       - Each round can have ncut = 1 - 61 using encoded character `[1-9,a-z,A-Z]`
       - If h-refine is off, it puts `0000`.
    - Read a file
      - Only used when simulation has h-refine
      - **Backward support** older format: empty spaces means `0000`.
      - Compare schedule from file and from simulation with a quick schedule check.
      - Once it accepts, it makes the diff to apply the needed refinement from checkpoint file to match the simulation. 
    - Requirement:
      - The "Base mesh" (before any refinement) for checkpoint and simulation must match.
      - If user call gen_re2 and run simulation with new re2, previous checkpoint files cannot be used. To bypass this, one need to zero out `nhref` before dumping file to clear it's restart stage..
  This replace the previous syntax below.

3. (old, removed) restart syntax
    Syntax: `startFrom = "r.fld href=2;3"`   
    Algo:
    a. Read original checkpoint file as-is   
    b. Modify the element mapping `er(i)` so the read elements are distributed to the correct MPI ranks.
    c. The read and distributed elements are then refined afterwards. 

**New Variables**
- `SIZE.inc`
  `lhref=10`: maximum number of refinement rounds

- `INPUT`
  `nhref`: rounds of refinement  
  `hrefcuts(lhref)`: refinement schedule

- `RESTART` (postfix `rs`)
  `nhrefrs`: rounds of refinement for a checkpoint file
  `hrefcutsrs(lhref)`: refinement schedule for a checkpoint file

**CI Tests**
- Add CI test ether/`EthierRefine`
  ```
  baseline
  hrefine = 2, dt = 1/2x, nstep = 2x
  hrefine = 3, dt = 1/3x, nstep = 3x
  hrefine = 2,2, dt = 1/4x, nstep = 4x
  hrefine = 4, dt = 1/4x, nstep = 4x
  ```

- Add CI tests into `io_test` for restart options
  - Original io_test (PnPn-2)
  - Extra hrefine tests with restart options: `href=2`, `href=3` and `href=2;2`
    - read io_test0.f00001 (fp32, normal)
    - read io_test0.f00002 (fp32, setics + `U S1 TIME=5.0`)
    - read io_test0.f00003 (fp64, if_full_pres=T)
      Example output from CI:
      ```
      PASSED:field coord x diff l2 norm =   0.3102081E-15
      PASSED:field coord y diff l2 norm =   0.1340270E-15
      PASSED:field coord z diff l2 norm =   0.1378978E-15
      PASSED:field vel x   diff l2 norm =   0.8872781E-04  (interp. err. at N=5)
      PASSED:field vel y   diff l2 norm =   0.1562231E-02
      PASSED:field vel z   diff l2 norm =   0.5401994E-04
      PASSED:field temp    diff l2 norm =   0.3825670E-04
      PASSED:field psc01   diff l2 norm =   0.1473877E-03
      PASSED:field psc02   diff l2 norm =   0.4018980E-03
      ```
  - Extra hrefine tests to let `hrefine=2,2` read from `hrefine=2`
  - Original io_test_rs for full restart

**Extra Notes**
- `ifaxis=T` is not supported. It needs more if statements for interpolation.
- restart: When there is `if_full_pres`, it skips reading pressure.
- NekDoc (https://github.com/Nek5000/NekDoc/pull/153) (Need update for the restart header)

Initial version is done by Paul @fischer1


yslan:
@stgeke @dshaver14 This PR is ready to review. Thanks.

PR #869, opened 2025-10-22
PnPn-2 + restart : fix pressure field

bbastien-cn:
Using the PnPn-2 formulation, the subroutine now interpolate the pressure field onto the velocity mesh instead of copying it. It prevents zeros values on GLL points to be introduced in the simulation .


yslan:
This behavior is by design. `full_restart` use `if_full_pres=T` to avoid extra interpolation and preserve all digits.

PR #874, opened 2025-12-07
Add validation checks in read_re2_hdr function

yslan:
The simulations will likely die at the later the checks but this helps detect error at an earlier location.


yslan:
Error likes this will be detected immediately and the message would be less confusing.
https://groups.google.com/g/nek5000/c/UESP_eXmQoQ

PR #875, opened 2025-12-08
Fix typo: vr_ifaxis -> chk_axis in drive1.f

ricardofrantz:
Commit f99f4b5 introduced a call to vr_ifaxis in drive1.f:113, but the subroutine is defined as chk_axis in bdry.f:2235. This causes undefined symbol errors when building with ifaxis=true.


yslan:
@fischer1

yslan:
Looks good to me. Recovers 16/26 CI tests before f99f4b5af81e8e561cd34e13e7a70d2d799f447e

PR #881, opened 2025-12-12
Allow initial bad mesh without abort

### LINKED ISSUE #646: Invalid Jacobian = Abort flag
Nek checks the mesh Jacobian for negative or 0 values at several stages.  For most scenarios, only one such check is worthy of aborting the run, that is just prior to execution of the first timestep.

This is particularly true if a user is loading a mesh in the restart file --- the earlier meshes generated from the .rea files are not used.

I propose a flag, ifjac0_abort, which when .true. indicates to Nek to stop the run, otherwise simply make a note to the logfile.

We could add a call to chkjac just after the first call to userchk (In case the user is modifying the geometry there) and, in that case, set ifjac0_abort=.true.   In all other cases
keep it .false., but allow the user to change it, for example, in userdat2 or at any other hook.

### LINKED ISSUE #762: Print mesh metrics after reading a mesh from the restart file
Current mesh metrics are printed after userdat2, which might not be the mesh used in the simulation.
We might also want to call mesh_metrics (needs xm1 to xc first) after reading a mesh from the restart file.

This is a similar issue to #646, and should be fixed together.

### LINKED ISSUE #877: Code hangs with negative Jacobians
The code sometimes hangs when negative Jacobians are detected on some ranks, but not on all ranks. It looks like this is caused by a call to `exitt` inside an `if` statement in `verrhe`.

https://github.com/Nek5000/Nek5000/blob/master/core/connect1.f#L911
https://github.com/Nek5000/Nek5000/blob/master/core/connect1.f#L943

ideally, it should always exit cleanly. It seems the logic is already there if these two exit statements are removed.

This may have some overlap with #646

yslan:
This PR adds an "allow bad mesh" mode so that users can inspect and possibly fix a bad mesh in `usrdat2`. Nek will now always enter `usrdat2` without exiting due to an initial mesh error.

With `ifjac0_abort = .false.`, the code continues at least until the first `userchk` call, so users can also perform meshing there or load a fixed mesh from a restart file.

Change log:
- Introduce `ifjac0_abort` to control whether the code aborts when a bad initial mesh is detected.
  - Defaults to `.true.`.
  - Can be overridden in `usrdat0`, `usrdat`, or `usrdat2`.
  - This enables a “pre-processing mode” in which users can fix the mesh in `usrdat2`, `userchk`, or by loading a checkpoint.
- Change the initial mesh check so it does not abort before entering `usrdat2`.
- Introduce a high-level `mesh_check` subroutine as a unified interface:
  - Includes both right-handedness and negative-Jacobian checks.
  - Supports configurable abort behavior on error.
  - Provides 3 verbosity levels for summary output.
  - Provides 3 debug modes when bad elements are present:
    - Can dump error type into `pr` (0=ok, 1=rhs, 2=jac, 3=both) and Jacobian into `temp`.

Default behavior (no user changes) is documented here:
https://github.com/Nek5000/Nek5000/pull/881#issuecomment-3648852204

This closes:
- https://github.com/Nek5000/Nek5000/issues/646
- https://github.com/Nek5000/Nek5000/issues/762
- https://github.com/Nek5000/Nek5000/issues/877

### DIFF
--- a/core/INPUT
+++ b/core/INPUT
@@ -51,7 +51,7 @@ c
      $               ,ifxyo_,ifaziv,ifneknek,ifneknekm,ifneknekc
      $               ,ifcvfld(ldimt1),ifdp0dt
      $               ,ifmpiio,ifcrrs,ifrich,ifvvisp
-     $               ,ifbmap(ldimt1)
+     $               ,ifbmap(ldimt1),ifjac0_abort
 
       common /input3/ if3d,ifflow,ifheat,iftran,ifaxis,ifstrs,ifsplit
      $               ,ifmgrid 
@@ -67,7 +67,7 @@ c
      $               ,ifcyclic,ifmoab,ifcoup, ifvcoup, ifusermv,ifreguo
      $               ,ifxyo_,ifaziv,ifneknek,ifneknekm,ifneknekc
      $               ,ifcvfld,ifdp0dt
-     $               ,ifmpiio,ifcrrs,ifrich,ifvvisp,ifbmap
+     $               ,ifmpiio,ifcrrs,ifrich,ifvvisp,ifbmap,ifjac0_abort
 
       logical         ifnav
       equivalence    (ifnav, ifadvc(1))
--- a/core/coef.f
+++ b/core/coef.f
@@ -530,7 +530,6 @@ C
       ENDIF
 C
       kerr = 0
-      ifprint = .true.
       DO 400 ie=1,NELT
          CALL MAP31 (RXM1(1,1,1,ie),RXM3(1,1,1,ie),ie)
          CALL MAP31 (RYM1(1,1,1,ie),RYM3(1,1,1,ie),ie)
--- a/core/coef.f
+++ b/core/coef.f
@@ -529,7 +529,6 @@ C
          CALL RONE  (TZM1,NTOT1)
       ENDIF
 C
-      kerr = 0
       DO 400 ie=1,NELT
          CALL MAP31 (RXM1(1,1,1,ie),RXM3(1,1,1,ie),ie)
          CALL MAP31 (RYM1(1,1,1,ie),RYM3(1,1,1,ie),ie)
--- a/core/coef.f
+++ b/core/coef.f
@@ -547,7 +547,7 @@ C
          CALL MAP31 (ZM1(1,1,1,ie),ZM3(1,1,1,ie),ie)
  400  CONTINUE
 
-      call mesh_check(ifjac0_abort,1)
+      call mesh_check(.false.,0,0)
       call invers2(jacmi,jacm1,ntot1)
 
       RETURN
@@ -625,7 +625,7 @@ C
          CALL ASCOL5  (TZM1,XRM1,YSM1,XSM1,YRM1,NTOT1)
       ENDIF
 C
-      call mesh_check(.false.,0)
+      call mesh_check(.false.,0,0)
       call invers2(jacmi,jacm1,ntot1)
 
       RETURN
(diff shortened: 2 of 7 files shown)


dshaver14:
~The logic seems to be off for this. Isn't the intent of #646 to allow setup to proceed through `usrdat2`? this PR seems to still quit with invalid elements from an `re2` file.~

wait, I misunderstood. 
I don't see the utility in allowing `ifjac0_abort` to be a user flag. Rather, we should allow the code to proceed through `usrdat2` regardless, then if the mesh is still bad, the code aborts.

fischer1:
I agree with Dillon's idea...  We don't fail till _after_ usrdat2.

We should check, however, both before and after --- because we want the user to know if it was something in the original mesh or in the post-processed mesh. ?

yslan:
Let me see if I understand correctly:
- No matter what, we want the code to run until usrdat2
- If user insists (ifjac0_abort=F), let the code run until userchk

If that's the case, we will probably check the mesh at three stages?
```
stage 1: re2, as-is
chk 1: default mesh check, but don't quit

stage 2. usrdat2
chk 2: extra check. (default) quit if failed. Allow continue if user wants.

stage 3. checkpoint or userchk
chk 3: extra check. Always quit if mesh failed.
```

re2 is not easy to open, so I'd at least want to see whether the original mesh fails or not
The other two checks are required because of the abort.

dshaver14:
I would say we can stop after `chk 2`. We should establish that `ursdat2` is the place to fix/change your mesh and by `usrdat3` it's too late. 

The checkpoint check is only if coordinates have been read from a restart file, which we should have been doing anyway.

yslan:
Sometimes, we project mesh by solving Laplacian. I think the best place to put it is inside the first usrchk?

We can also control the verbose level so it only prints when failed. Or, save mesh metrics and only print when it differs.

dshaver14:
as long as we discussing it, it would be very helpful if the xyz file that gets dumped for vanishing Jacobians contained the Jacobians as a scalar to aid in visualization.

yslan:
Now the code does this:
```
(usrdat0)
stage 1: re2, as-is
  chk1a: default right-handiness check, quit when it found error, but can be overwrite.
(usrdat)
  chk1b: default jacm3 check, print first bad elem per rank, dump xyz with jac, but don't quit
  chk1c: default jacm1 check, print first bad elem per rank, dump xyz with jac, but don't quit

stage 2: usrdat2
  chk 2: extra check. (default) quit if failed. Allow continue if user wants, print mesh metrics (as before)

stage 3: checkpoint or userchk
  chk 3: extra check. Always quit if mesh failed. compute mesh metrics but only prints if they have changed since usrdat2. 
```

TODOs:
- [x] mesh 3. There is an extra mesh check for temporary xm3 and jacm3. I'm not sure what the mesh3 is for. It looks like a temporary one that is used to setup xm1? This seems redundant to me.
- [x] right-handiness. What do we use this for? connectivity and periodicity? Should we check it again later?
I'd like to follow the same no-abort-until-usrdat2 logic, but need to make sure it's not something irreversible.

dshaver14:
For right-handedness, the code checks the cross product of the two r-s edge vectors for each element vertex (or the cross product of the r-s edge vectors dotted with the t-normal edge vector in 3D). It's done before the GLL points are laid out and ignores midpoints, connectivity, and periodicity. I don't think it needs to be repeated, but it's useful to do it once since it give a little more information about where the issue lies.

@yslan I've addressed the logic I'd like to see in the PR I made to your msh_chk branch. I really just don't want it to quit here because I can't get any output on which elements are bad like I can with the xyz dump. Note that if it fails here, it will definitely fail in the Jacobian check.

yslan:
Confirm that mesh 3 is never used, reuse `mesh_check` on mesh 1 in `glmapm3`. (lx3=lx1, map31 interpolate is identity).

Turn right handiness into warning rather than error that require exit as the error will be detected by negative Jacobian later.

I think this PR is ready.

dshaver14:
The printout logic in `chkjac` could be better. Right now, it prints out the first negative Jacobian it finds per rank. I think we should only print out the total number of non right-handed elements and the total number of negative Jacobians by default and hide the coordinate printout of all elements/nodes behind a higher loglevel. Thoughts?

yslan:
@dshaver14 I like the idea of putting rhs check into `mesh_check` and make it a high level subroutine that user can even use it inside `usrdat2`.

This "allow-bad-mesh" mode is for large meshes. Otherwise, there are prenek/postnek to visualize the location. Therefore, I agree dumping coordinates is not particularly useful.

Then, I propose to add `idbg` as debugging option and only dump bad elements after userchk?
```
c     When bad elements detected, what to do?
c     idbg    : 0          do nothing
c             : 1          dump xyz file, jacobian, and error type (e.g., 0=ok, 1=rhs, 2=jac, 3=both)
```

Since we are here, should we use new format p66=6? or p66=4 is good enough?

dshaver14:
yea, I like this. Can I also suggest an option to dump only the bad elements? If the jacobian is very close to zero, the threshold operation in ParaView struggles.

I think we shouldn't force the format. We can leave p66 as is and let the user change it in usrdat if they want.

yslan:
If we assign a number into ~vx~ pr, user can set threshold: 
```
bad_elem(ie),  0=valid, 1=LHS, 2=neg-jac, 3=both.

0.5 < pr < 1.5: LHS elements (excluding neg-jac)
1.5 < pr: all neg-jac elements
2.5 < pr < 3.5: both neg-jac and LHS elements
0.5 < pr: bad elements
```

> dump only the bad elements? 

With the err type saved into pr, I assume the leftover usecase is to save IO and Viz cost at scale. However, when the bad location is sparse or scattered at many places, this is not always useful.
Personally, I sometimes also dump the neighbors of bad elements for a better visualization. User can do this manually in usrdat2.

yslan:
It turns out the entire `SCRCT` is a scratch memory.
`xyz`, declared in the common block `ctmp1`, share address with other location using `ctmp1`. To avoid conflict, we make `xyz` a local variable, and pass it to argument.

Now, it's ready for another round of review.

dshaver14 (approved):
I've tested it and it looks like it's ready to merge.

Issue #873, opened 2025-12-02
HYPRE doesn't build with CMake 4.2.0


khanhn201:
Workaround: after the install script extracted the HYPRE tar file, change the minimum version in 3rd_party/hypre/src/CMakeLists.txt
```
cmake_minimum_required (VERSION 3.5.0)
```

PR #882, opened 2025-12-12
3rd party versioning cleanup; remove parMETIS

### LINKED ISSUE #825: parRSB/gslib fail to update properly
`$makenek clean` doesn't delete the unpacked versions of parRSB and gslib. When parRSB, for example, is updated, the user needs to manually delete the `Nek5000/3rd_party/parRSB/parRSB` directory for the new version to be downloaded and installed. The `install` script only checks for the existence of the tool subdirectory when deciding if it needs to download a new tarball, not whether it is the correct version. We could tag the subdirectory with the version number to make sure it's always consistent with the version of Nek. Something like:

```
#!/bin/bash
set -e

VER=0.9.2

if [ "$1" == "clean" ]; then
  rm -rf parRSB_v$VER/build lib include 2>/dev/null
  exit 0
fi

if [ -f ./lib/libparRSB.a ]; then
  exit 0
fi

if [ ! -d ./parRSB_v$VER ]; then
  rm -rf ./parRSB* 2>/dev/null #delete any version that might already exist
  wget --no-check-certificate -O v$VER.tar.gz https://github.com/Nek5000/parRSB/archive/v$VER.tar.gz
  mkdir parRSB_v$VER
  tar -zxvf v$VER.tar.gz -C ./parRSB_v$VER --strip-components=1
fi

cd parRSB_v$VER
set -x
make lib install $PARRSB_OPT
set +x
```

yslan:
- Introduce a "`distclean`" mode that removes both tarballs and extracted source directories.
  The existing "`clean`" mode now only removes the build and install artifacts while keeping the source directory and tarball.
- Extract third-party sources into versioned directories (e.g., `gslib_v1.0.9`) while maintaining a symlink `gslib -> gslib_v1.0.9`) so existing include path remain valid.
- Introduce an `install.stamp` file to record the version and timestamp of the last successful install. If version mismatched, a `clean` will be performed prior to the build.
- Introduce a "`fetch`" mode to purely download tar files. This is for `nekdist`, not for `makenek`.
- Unify the interface and behavior of all third-party installs.
- Tract the changes of install script
- Remove parMETIS

Users can now unpack a third-party tarball, modify the sources, and build Nek cases against the modified library. Those changes are preserved unless `makenek distclean` is run.

After a `git fetch; git pull`, running `makenek eddy` again will see the version change via `install.stamp`, trigger a `clean`, and rebuild the third-party library automatically.

Not a solution of https://github.com/Nek5000/Nek5000/issues/873 but will make modification easier (once per clone, unless `distclean`)
Close https://github.com/Nek5000/Nek5000/issues/825


yslan:
TODOs:

- [x] @thilinarmtb parMETIS link seems to be broken. Need verification that the patches are still working for the new link.

- [x] integrate with `nekdist`
- [x] CI coverage of using all libs
  - [x] gslib and blasLapack are used all the time
  - [x] HYPRE tested in `Ethier::test_PnPn_Parallel`
  - [x] CVODE tested in `MvCylCvode::test_PnPn_Parallel` and `ConjHt::test_PnPn_Parallel`
  - [x] PARRSB, co2+RSB tested in `Ethier::test_PnPn_Parallel` along with HYPRE.
  - [x] ~PARMETIS?~

yslan:
All 3rd party libraries are covered by CI tests. This PR is ready for review.

dshaver14:
@yslan , can you update all `wget` urls to `https`? This makes a difference on our internal clusters.

PR #884, opened 2025-12-29
Add LLVM flang support

stgeke:


core/convect2.f
yslan, line 1: It looks like the compiler doesn't check the index if it's not a compile time constant.
We could also use `i0 = 0` and preserve original index `i0 + 9`. After all, compiler won't know whether i0 is non-negative.

PR #903, opened 2026-06-03
Avoid reserved Fortran I/O unit 101

yslan:
Per the HPE/Cray Fortran documentation [here](https://cpe.ext.hpe.com/docs/24.11/getting_started/CPE-CCE-Fortran.html#file-unit-numbers), Cray Fortran 20 reserves Fortran units `100`, `101`, and `102` for `stdin`, `stdout`, and `stderr`, respectively, through `ISO_FORTRAN_ENV`.

Nek5000 currently uses `unit=101` when generating the `.nek5000` file. On ALCF/Polaris and ALCF/Crux (using PrgEnv-Cray), it will show the error:
```text
lib-4013 : UNRECOVERABLE library error
  The unit number (101) in the OPEN statement must be nonnegative and outside
  the range of 100 through 102.

Encountered during an OPEN of unit 101
Fortran unit 101 is connected to a sequential formatted text file
  (standard output).
```

This PR simply avoid it with a different unused number.

Found by Ramesh.

### DIFF
--- a/core/prepost.f
+++ b/core/prepost.f
@@ -1223,11 +1223,11 @@ c-----------------------------------------------------------------------
 
       if(nid.eq.0) then
         call chcopy(nam1(kk),'.nek5000',8)
-        open(unit=101,file=name)
-        write(101,*) "filetemplate: ", trim(fname), "%01d.f%05d"
-        write(101,*) "firsttimestep: 1"
-        write(101,*) "numtimesteps: ", nfld
-        close(101) 
+        open(unit=89,file=name)
+        write(89,*) "filetemplate: ", trim(fname), "%01d.f%05d"
+        write(89,*) "firsttimestep: 1"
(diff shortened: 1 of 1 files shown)


thickAsABrick:
Thanks for the fix, Yu-Hsiang! Made the following change in prepost.f:
---
1227 c..        open(unit=101,file=name)
1228         open(unit=201,file=name)
1229         write(201,*) "filetemplate: ", trim(fname), "%01d.f%05d"
1230         write(201,*) "firsttimestep: 1"
1231         write(201,*) "numtimesteps: ", nfld
1232         close(201)
---
and it worked!

yslan:
Although it's less likely to happen, but I think putting number below 100 would be safer in case they reserve a different number in the future. That's why I choose 89 in the PR.

Anyway. CI tests are passed and the code now worked on Crux. This PR is good to go.

PR #904, opened 2026-06-03
Modified code to not use unit=101, which is a reserved I/O unit in th…

thickAsABrick:
…e Cray-Fortran compiler


yslan:
duplicate of https://github.com/Nek5000/Nek5000/pull/903

PR #910, opened 2026-07-28
mfi refactor for a batched read assign

yslan:
This PR refactor mfi to penetrate the batches to read, redistribution, and assign.
- Allow reading data that is larger than buffer.
- Allow capping max elements per batch to ease communication pressure
- Clean up RMA and CR paths.
- Clean up CI. All 7 IO_Tests cases are passed. (See short_tests/io_test/README.md)

## Buffers
In words.
- `wk`: a unified buffer for RMA and CR
- `w2`: a read buffer holds one-file-batch data.

## Overall Structure / Design
```
do batch
  MPIIO read
  mfi_redist_plan
  do round
    mfi_redist_round_[cr,rma]                                                                                
    mfi_assign_[cr,rma]                                               
  enddo   
enddo
```

- Each file is read in **batches** of `nbe` elements, sized by `lbrst` (user input) or what fits `w2` (depends on `nxyzr`).
- To avoid incoming data overflow recv buffer, redistribution is split into bounded **rounds** which is a subset of batch.

`mfi_redist_plan` is used to compress index map into CSR + handshake to determine what element to send in this round. This avoid looping over all local element and having multiple small (near empty) rounds.

Co-Authored-By: Claude Opus 4.8 (1M context) noreply@anthropic.com and reviewed by Codex.


core/modules/VRTHOV.f
```diff
@@ -1,32 +1,36 @@
...
          implicit none
 
-         integer ierr, nvrthov, nvrthov_i
+         integer ierr, nvrthov
 c        ic.f (mfi_gets/mfi_getv: w2(lrbs), lrbs=20*lx1*ly1*lz1*lelt)
          integer sz_w2
-c        ic.f (mfi_gets/mfi_getv: vi(2+lrbs_loc,lelt))
-         integer sz_vi
 
          sz_w2 = 20*lx1*ly1*lz1*lelt
```
stgeke: Why do we still need this larger buffer size? That's something we should avoid.

yslan: It's the same. Before:
```
      parameter(lrbs_loc=20*lx1*ly1*lz1)
      parameter(lrbs=lrbs_loc*lelt)
      common /vrthov/ w2(lrbs) ! read buffer
```

stgeke: Sure, but with a batched implementation, we should be able to use a much smaller buffer.

yslan: I tested in reverse way by reading a larger file using the same buffer.
Small buffer should work but I don't know what value is good and not sure if it will slow down the code.

Let me test it. Any suggestions on expected buffer size?

stgeke: I think we can safely use `lrbs_loc = lx1*ly1*lz1`.

yslan: Actually, we still use element as an unit of payload.
So, `lrbc_loc` should hold (lx1+6)^3 anticipate a N+6 -> N interpolation, which is the next limit inside mapab.

Dropping lelt seems to be possible.

yslan:
New default: 16MB buffer for both read (w2) and redist (wk),
constant in problem size. Replace `init()` with `reserve()` mode, called inside mfi. Grows only if one element exceeds 16MB.

Two knobs (element counts; reduce-only usage of the fixed 16MB buffer):
```
nrst_rd: read batch -- #elements read into rdbuf per pass (IO dial)
    0 (default) auto = fill the buffer: nbe = min(16MB/nxyzr, mtup)
    >0          cap at min(nrst_rd, lelt)
                larger  -> fewer/larger reads (better IO)
                smaller -> smaller reads + smaller per-batch handshake
nrst_cm: redist round -- #payloads per all-to-all (network dial) into cmbuf
    1024 (default) good CR comm size; cap = min(1024, mtup)
    0          auto = fill the buffer: cap = mtup (~16MB / one element)
    >0          cap = min(nrst_cm, mtup)
                smaller -> smaller all-to-alls, lower peak network traffic
```

PR #913, opened 2026-08-10
Update postpro.f cubic function

sierratutwiler:
Removed absolute value from line 227 to correct eigenvalue sign inversion issue


yslan:
link https://github.com/Nek5000/Nek5000/issues/912

core/postpro.f
yslan: Since we are here, can you also fix this line?
It should be `xo`, not `x`

yslan:
Confirm the fix. Here is a test:

https://godbolt.org/z/4TWcrEh6T

sierratutwiler:
Checked the fix through the test, the new version looks like it is passing.

khanhn201:
Left is the fix in this pr, right is the old lambda2, both at the same lambda2 level.
This run I have is a bit too under-resolved but in general the fix seems to clean up the lambda2.

a0
a1
a2
a3

fischer1:
Fantastic...Apologies to all for this error.  I notice that it is also in postnek, in postnek3.f

yslan:
Merge it as it passes the random root tests and improve the isosurface slightly. Close #912

Good finding @sierratutwiler!  
@khanhn201 thanks for testing.

